PR #11, opened 2023-07-18
Add AD type for FiniteDifferences

devmotion:
Maybe it would be reasonable to add a type for FiniteDifferences as well?

A bit inconsistent, in contrast to #10 in this PR I opted for a default value of `nothing` since I wanted to avoid pulling in FiniteDifferences which has much more dependencies than EnzymeCore. Alternatively, on Julia >= 1.9 one could add a weak dependency on FiniteDifferences and define a constructor with a more concrete default value such as `FiniteDifferences.central_fdm(5, 1)` (but this would lead to inconsistencies between Julia < 1.9 and >= 1.9...).


Vaibhavdixit02:
We should avoid that as you noted, for the extension we can do it separately but I think it might even be sufficient as it is now and the package using it can work around it

PR #15, opened 2023-08-11
Use more abstract types to create a hierarchy

Vaibhavdixit02:


src/ADTypes.jl
```diff
@@ -16,37 +25,37 @@ function AutoFiniteDiff(; fdtype = Val(:forward), fdjtype = fdtype,
...
 
-struct AutoReverseDiff <: AbstractADType
+struct AutoReverseDiff <: AbstractReverseMode
     compile::Bool
 end
 
 AutoReverseDiff(; compile = false) = AutoReverseDiff(compile)
 
-struct AutoZygote <: AbstractADType end
+struct AutoZygote <: AbstractReverseMode end
```
avik-pal, line 48:
```suggestion
struct AutoZygote <: AbstractReverseMode end

struct AutoSparseZygote <: AbstractSparseReverseMode end
```

avik-pal: Let's just add this because that is what triggered the whole discussion

avik-pal (approved):

src/ADTypes.jl
```diff
@@ -16,37 +25,39 @@ function AutoFiniteDiff(; fdtype = Val(:forward), fdjtype = fdtype,
...
+
+struct AutoSparseZygote <: AbstractSparseReverseMode end
 
 struct AutoEnzyme{M} <: AbstractADType
     mode::M
 end
 
 AutoEnzyme(; mode = nothing) = AutoEnzyme(mode)
```
avik-pal, line 56: Should we split Enzyme into forward and reverse modes to fit into this better?

Vaibhavdixit02: I was conflicted about this as well. The `mode` field should take the choice per the current definition. That can be the Enzyme `Forward` and `Reverse` objects directly, it's `nothing` because a default mode doesn't make sense to me it would be completely up the package if they want to put a default mode.

src/ADTypes.jl
```diff
@@ -16,37 +25,39 @@ function AutoFiniteDiff(; fdtype = Val(:forward), fdjtype = fdtype,
...
 end
 
 AutoReverseDiff(; compile = false) = AutoReverseDiff(compile)
 
-struct AutoZygote <: AbstractADType end
+struct AutoZygote <: AbstractReverseMode end
+
+struct AutoSparseZygote <: AbstractSparseReverseMode end
 
 struct AutoEnzyme{M} <: AbstractADType
```
Vaibhavdixit02, line 52:
```suggestion
struct AutoEnzyme{M} <: Union{AutoForwardMode, AutoReverseMode, AutoSparseForwardMode, AutoSparseReverseMode}
```

Vaibhavdixit02: This could be an option 😅

PR #19, opened 2023-11-14
Drop support for Julia < 1.6

prbzrg:
We did it for CI in #13, but `compat` should be updated too.


Project.toml
```diff
@@ -4,7 +4,7 @@ authors = ["Vaibhav Dixit <vaibhavyashdixit@gmail.com> and contributors"]
 version = "0.2.4"
```
Vaibhavdixit02, line 4:
```suggestion
version = "0.2.5"
```

Issue #25, opened 2024-03-09
Documenter web page?


gdalle:
**Is your feature request related to a problem? Please describe.**

Since #23 we have docstrings, but they are not shown anywhere right now. As a result, the API specification is still imprecise for users, which is a problem when so much of SciML depends on this package.

**Describe the solution you’d like**

A Documenter.jl web page showcasing the docstrings and the list of exported names.

**Describe alternatives you’ve considered**

I tried adding a list of exports to the README in #23, but it will become periodically out of date (it already is since #21 was merged too). A Documenter workflow is more adaptive.

**Additional context**

I can set up the Documenter stuff by myself, but there are things I cannot do:

- Add the required secrets to the repo. If you want to give me access that's also fine.
- Integrate this to a wider SciML multidoc. Not sure we want to do that though, because I envision this package being foundational outside of SciML too thanks to https://github.com/gdalle/DifferentiationInterface.jl

Vaibhavdixit02:
I can set it up. But in case you had already done something for this I have added you to the repository.

gdalle:
Thanks! I can do it as well, once #26 is merged

Once there is a nice page up and running we can bump the patch version and release, if that's okay?

PR #26, opened 2024-03-10
Document AutoChainRules

gdalle:
- Add docstring to `AutoChainRules`
- Add `AutoChainRules` to README

### DIFF
--- a/src/ADTypes.jl
+++ b/src/ADTypes.jl
@@ -19,6 +19,15 @@ abstract type AbstractSparseReverseMode <: AbstractReverseMode end
 abstract type AbstractSparseForwardMode <: AbstractForwardMode end
 abstract type AbstractSparseFiniteDifferences <: AbstractFiniteDifferencesMode end
 
+"""
+    AutoChainRules{RC}
+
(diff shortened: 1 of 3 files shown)


gdalle:
Once this is merged, would it be possible to tag a new release?

src/ADTypes.jl
```diff
@@ -19,6 +19,15 @@ abstract type AbstractSparseReverseMode <: AbstractReverseMode end
...
+    AutoChainRules{RC}
+
+Chooses any AD library based on [ChainRulesCore.jl](https://github.com/JuliaDiff/ChainRulesCore.jl), given an appropriate [`RuleConfig`](https://juliadiff.org/ChainRulesCore.jl/stable/rule_author/superpowers/ruleconfig.html) object.
+
+# Fields
+
+- `ruleconfig::RC`
+"""
 Base.@kwdef struct AutoChainRules{RC} <: AbstractADType
     ruleconfig::RC
```
Vaibhavdixit02, line 32: Can this have a default value?

Vaibhavdixit02:
Feel free to bump the version in this PR itself

src/ADTypes.jl
gdalle: Not without loading an extension

gdalle:
> Feel free to bump the version in this PR itself

Done

gdalle:
But we should probably wait until I address #25 before releasing 0.2.7, that way we will have a nice documentation too

src/ADTypes.jl
Vaibhavdixit02: Okay, makes sense

Issue #38, opened 2024-04-10
Change handling of sparse backends


gdalle:
Problems:

- The code for `AutoSparseBackend` is a duplicate of `AutoBackend`, sometimes with misses (see #33)
- There is no built-in way to specify additional aspects like the sparsity detection method or the coloring algorithm
- It is tedious to enumerate all possible sparse backends

Suggested solutions (instead of #37):

- Replace `AutoSparseBackend` with a wrapper struct `AutoSparse(AutoBackend)`
- Add fields to this wrapper struct for sparse-specific ingredients

```julia
struct AutoSparse{B<:AbstractADType,S,C} <: AbstractADType
    backend::B
    sparsity_detector::S
    coloring_algorithm::C
end
```

To preserve backwards-compatibility we can try to define shortcuts like the following, to be removed in v0.3.0:

```julia
const AutoSparseBackend = AutoSparse{<:AutoBackend}

function AutoSparseBackend(args...; sparsity_detector, coloring_algorithm, kwargs...)
    backend = AutoBackend(args...; kwargs...)
    return AutoSparse(backend, sparsity_detector, coloring_algorithm)
end
```

However I'm not 100% sure we can keep this from being breaking.

Vaibhavdixit02:
Definitely agree that the current way of AutoSparse* is pretty tedious and not modular.

I'd keep the name something else since `AutoSparseBackend` would make it appear like an AD backend so just `Sparse(args...)` seems nicer?

gdalle:
I thought about that but `Sparse` is a bit generic, hence the addition of `Auto`. Besides, we would actually have
```julia
AutoSparse <: AbstractADType
```
so it seems coherent.

Vaibhavdixit02:
I see, that seems reasonable

gdalle:
@avik-pal @ChrisRackauckas  any thoughts?

Vaibhavdixit02:
One more option AutoSparsifyBackend for the constructor? 😅

gdalle:
>  just Sparse(args...) seems nicer?

The killer argument is that we will export that name, and we cannot decently export something named `Sparse`

> One more option AutoSparsifyBackend for the constructor? 😅

What do you mean?

ChrisRackauckas:
AutoSparse <: AbstractADType is a very good middle ground IMO.

[...]

Issue #24, opened 2024-03-08
What is the right way to specify the Enzyme mode?


gdalle:
**Question❓**

The source file says nothing, the test file says something but does something else:
```julia
# In practice, you would rather specify a
# `mode::Enzyme.Mode`, e.g. `Enzyme.Reverse` or `Enzyme.Forward`
adtype = AutoEnzyme(; mode = Val(:Reverse))
@test adtype isa ADTypes.AbstractADType
@test adtype isa AutoEnzyme{Val{:Reverse}}
```
What convention is used in the SciML ecosystem? Can it be documented?

Vaibhavdixit02:
I haven't been using the `mode` kwarg, idk if @avik-pal uses it. For me in Optimization.jl it is just reverse mode for gradient and forward over reverse for hessian so not well suited for the mode argument.

avik-pal:
Me neither I kind of always assume it's reverse mode (which I probably shouldn't)

gdalle:
Another question is what to put inside the AutoEnzyme object, beyond the "mode". Do we need more information to make informed decisions?

Vaibhavdixit02:
>Since you're maintaining packages that make use of ADTypes.jl, I'd be glad to have your opinion on https://github.com/gdalle/DifferentiationInterface.jl.

It looks great! Some other features already being discussed in https://github.com/gdalle/DifferentiationInterface.jl/issues/26, caching etc and support for sparsity will also be pretty essential for me to consider using it in Optimization.jl

gdalle:
@avik-pal There's also another issue with using Enzyme.Reverse as mode, which is that depending on the function (like "value_and_gradient") you may want Enzyme.ReverseWithPrimal instead

gdalle:
@Vaibhavdixit02 indeed! Cache support is fully there for ForwardDiff and ReverseDiff (with and without compiled tape) already, with a unified interface. FiniteDiff and SparseDiffTools are next on my hit list

avik-pal:
> There's also another issue with using Enzyme.Reverse as mode, which is that depending on the function (like "value_and_gradient") you may want Enzyme.ReverseWithPrimal instead

Agreed

PR #40, opened 2024-04-12
Restructure sparse backends and replace subtyping by traits

gdalle:
## Motivation

This is a breaking PR (bumping version to v0.3.0) that restructures the handling of sparse backends, and cleans up the package in the process.

> TLDR: Most (concrete) structs have not changed too much, but
> - their abstract subtypes are replaced by a `mode` trait
> - their sparse counterparts are replaced by the `AutoSparse` struct

## Breaking API changes

- Bump Julia compat lower bound to 1.10

**Mode**

- Remove the abstract supertypes for the different modes (e.g. `AbstractForwardMode` and `AbstractSparseForwardMode`), and replace them with the `mode` trait. They were not documented but some packages used them anyway, like DifferentiationInterface.jl.
- Make `mode(ad::AbstractADType)` return one of the following singletons: `ForwardMode()`, `ReverseMode()`, `SymbolicMode()` or the ambiguous `ForwardOrReverseMode()` (for cases like `AutoEnzyme` and `AutoChainRules`)
- Assign the finite differences packages to `ForwardMode`

**Concrete structs**

- Remove `AutoModelingToolkit`
- Document the use of the `mode` field for `AutoEnzyme`: either `nothing` or a subtype of `EnzymeCore.Mode` (see https://github.com/SciML/ADTypes.jl/issues/24)
- Remove the `fdm = nothing` default for `AutoFiniteDifferences`, since it always has to be a finite difference method
- Add a tag and the associated type parameter to `AutoPolyesterForwardDiff` (see https://github.com/JuliaDiff/PolyesterForwardDiff.jl/issues/18)
- Remove the `AutoSparseForwardDiff` type and others like it, and turn them into deprecated constructors:
```julia
@deprecate AutoSparseForwardDiff(; kwargs...) AutoSparse(AutoForwardDiff(; kwargs...))
```

## Non-breaking API changes

- Export `AbstractADType`, which is useful for many downstream users
- Document keyword argument constructors (part of the API now)
- Add `AutoSymbolics` to replace `AutoModelingToolkit`
- Add `AutoTapir` for https://github.com/withbayes/Tapir.jl

**Sparsity**

- Introduce an `AutoSparse` struct that wraps a `dense_ad` backend with a `sparsity_detector` and a `coloring_algorithm` (as discussed in #38)
- Define getter functions for this new struct
- Define abstract types and default (trivial) choices for
  - the sparsity detector (`AbstractSparsityDetector`, `NoSparsityDetector`)
  - the coloring algorithm (`AbstractColoringAlgorithm`, `NoColoringAlgorithm`)
- Define utilities for getting Jacobian/Hessian sparsity patterns based on an `AbstractSparsityDetector`:
  - `jacobian_sparsity`
  - `hessian_sparsity`
- Define utilities for getting column/row colorings based on an `AbstractColoringAlgorithm`:
  - `column_coloring`
  - `row_coloring`

## Other changes

**Dependencies**

- Add ChainRulesCore.jl and EnzymeCore.jl as weak dependencies to define more specific `mode` dispatches
  - based on `ChainRulesCore.RuleConfig` for ChainRulesCore.jl
  - based on `EnzymeCore.Mode` for EnzymeCore.jl

**Package quality**

- Split source code into several files
- Add code coverage to CI (see https://github.com/SciML/ADTypes.jl/issues/41)
- Add more tests
- Include Aqua and JET to tests for code quality
- Add and improve docstrings
- Structure API reference in the docs with sections
- Clarify README

## Checklist

## Questions

- Should we document type parameters of the concrete structs? I feel like this limits us in the range of non-breaking changes we can do: for instance, turning `AutoPolyesterForwardDiff{chunksize}` into `AutoPolyesterForwardDiff{chunksize,T}` is breaking because someone might have used the former as a constructor. That is why I explicitly document keyword constructors and fields.
- Is the weak dependency handling okay on Julia < 1.9? The alternative would be to avoid defining the mode trait altogether, which might be preferable
- Should we pull in Diffractor.jl and Zygote.jl in the test suite to check on actual instances of the `RuleConfig` paradigm? I define very simple `RuleConfig`s in `runtests.jl` but it's less reliable than taking the real ones


src/abstract.jl
```diff
@@ -0,0 +1,36 @@
...
+Supertype for all AD choices.
+"""
+abstract type AbstractADType{mode} end
+
+"""
+    AbstractFiniteDifferencesMode
+
+Supertype for AD choices that rely on [finite differences](https://en.wikipedia.org/wiki/Finite_difference).
+"""
+const AbstractFiniteDifferencesMode = AbstractADType{:finite}
```
ChrisRackauckas, line 13: Why symbols here? Normally we would use an EnumX to stop comparison issues with spelling errors. But I don't quite get what this symbol is trying to do, is it supposed to specify a trait? Why not define a trait function then?

ChrisRackauckas: Or rather, why is this not actually an abstract type?

src/sparse.jl
```diff
@@ -0,0 +1,164 @@
...
+    AbstractColoringAlgorithm
+
+Abstract supertype for Jacobian/Hessian coloring algorithms, defined for example in [SparseDiffTools.jl](https://github.com/JuliaDiff/SparseDiffTools.jl).
+
+# Required methods
+
+- [`column_coloring`](@ref)
+- [`row_coloring`](@ref)
+"""
+abstract type AbstractColoringAlgorithm end
```
ChrisRackauckas, line 57: What's the plan for de-conflicting it with https://github.com/JuliaArrays/ArrayInterface.jl/blob/master/src/ArrayInterface.jl#L367 ?

src/sparse.jl
```diff
@@ -0,0 +1,164 @@
...
+Return the coloring algorithm for a sparse AD choice.
+
+# See also
+
+- [`AutoSparse`](@ref)
+- [`AbstractColoringAlgorithm`](@ref)
+"""
+coloring_algorithm(ad::AutoSparse) = ad.coloring_algorithm
+
+# TODO: document or remove these?
```
ChrisRackauckas, line 160: Can probably just remove. Parametric dispatching seems like it should be used to handle this, or trait functions.

ChrisRackauckas:
Something not mentioned in the top there is that this will require that ArrayInterface.jl take on an ADTypes.jl dependency, which is fine given that this is lightweight and all, but it should be noted that its `ColoringAlgorithm` needs a `@deprecate ColoringAlgorithm = ADTypes.AbstractColoringAlgorithm` which is then used in downstream implementation of its functions. `fast_matrix_colors` and `matrix_colors` live there since they are part of the extended sparse matrix interface, so that array types can define their coloring pattern overloads (things like BandedMatrices) without taking a full dependency on the AD libraries, but then specialize whenever such structured matrices are hit with the coloring algorithm (ex. BandedMatrices has a nice analytical solution based on the band size).

ChrisRackauckas:
The symbols instead of abstract types thing is weird. That needs some explanation.

src/abstract.jl
gdalle: The idea was that `AutoSparse(AutoForwardDiff())` should have the same supertype as `AutoForwardDiff()`, aka `AbstractForwardMode`. But it is impossible to do the following in Julia (see this [Slack exchange](https://julialang.slack.com/archives/C6A044SQH/p1712899380312649))
```julia
struct AutoSparse{AD} <: TheSpecificSuperTypeOfAD <: AbstractADType end
```
So the trick I found is to make `AbstractADType` parametric, and to redefine `AbstractForwardMode` and its friends as aliases.
I don't really care if we use `Symbol`s or an `Enum` here, I can change to an `Enum` if you think it's better. But that is the rationale anyway.

src/sparse.jl
gdalle: Yeah I had seen that, but it is a conscious proposition from me. Since ArrayInterface has way more dependencies than ADTypes, I didn't want to import the former here because ADTypes is gonna be at the root of hopefully the whole ecosystem.
Two solutions:
- (my least favorite) import ADTypes in ArrayInterface instead
- (my favorite) keep this code in ADTypes and remove it from ArrayInterface altogether, since it is very specific to differentiation

src/sparse.jl
gdalle: My aim in keeping them was to be as gently breaking as possible, but if you greenlight it I can remove them.
Similar question for the deprecation

gdalle:
Thanks for the review! Indeed you identified two parts where I was unsure:

> The symbols instead of abstract types thing is weird. That needs some explanation.

See my comment https://github.com/SciML/ADTypes.jl/pull/40#discussion_r1563914494

> will require that ArrayInterface.jl take on an ADTypes.jl dependency

Either that or the coloring names are dropped from ArrayInterface altogether? See my comment https://github.com/SciML/ADTypes.jl/pull/40#discussion_r1563915898

src/abstract.jl
ChrisRackauckas: What's the reason for not relying on trait functions for this information instead?

gdalle:
> its ColoringAlgorithm needs a @deprecate ColoringAlgorithm = ADTypes.AbstractColoringAlgorithm

I can rename `AbstactColoringAlgorithm` to minimize the refactoring work

> fast_matrix_colors and matrix_colors live there since they are part of the extended sparse matrix interface,

Not sure how to deal with those though. I can also rename `column_coloring` and `row_coloring` but these are different algorithms, and I don't see a similar toggle in `matrix_colors`?

src/sparse.jl
ChrisRackauckas: But then who defines the matrix coloring function definitions for base structured types?

src/abstract.jl
gdalle: If we want to be more ambitious (but more breakin) we can replace it all with a `mode` trait, which is less cumbersome than this subtyping trick. It might make sense for backends that are both forward- and reverse-compatible, like Enzyme. Internally, it is what DifferentiationInterface does anyway.

ChrisRackauckas:
> I can rename AbstactColoringAlgorithm to ColoringAlgorithm to minimize the refactoring work

It's at least contained to two repos, so it's not too big of a deal if it's deprecated properly. 

> Not sure how to deal with those though. I can also rename column_coloring and row_coloring but these are different algorithms, and I don't see a similar toggle in matrix_colors?

Yeah that's mostly by adjoint. But the bigger thing would be updating all of the downstream analytical solutions to have these functions as well.

src/sparse.jl
ChrisRackauckas: Go for it

src/sparse.jl
gdalle: Yeah I hadn't thought of that. Then I guess it's better if those names live in ADTypes and are imported by ArrayInterface? I don't know if matrix coloring has other prominent uses besides autodiff

gdalle:
> But the bigger thing would be updating all of the downstream analytical solutions to have these functions as well.

If we keep the name `matrix_colors` and import ADTypes into ArrayInterface, there is nothing to change at all. But that would mean having only one coloring function, and if I want the row coloring I take the column coloring of the transpose?

src/abstract.jl
gdalle: The reason for not using traits is that I wanted to do a soft PR that breaks as little as possible. But as you can see below, it is the nicer option, especially for backends who can do both forward and reverse mode:

https://github.com/gdalle/DifferentiationInterface.jl/blob/84d98a356a7e963b4b1acec09904edc78a70a033/DifferentiationInterface/src/traits.jl#L17-L20

https://github.com/gdalle/DifferentiationInterface.jl/blob/84d98a356a7e963b4b1acec09904edc78a70a033/DifferentiationInterface/ext/DifferentiationInterfaceEnzymeExt/DifferentiationInterfaceEnzymeExt.jl#L47-L48

ChrisRackauckas:
> But that would mean having only one coloring function, and if I want the row coloring I take the column coloring of the transpose?

That's what we currently do, but if you're going to setup general sparse AD then it's not a good idea since you won't easily know if someone passes a coloring pattern if it should be A or A'. So I like the idea of two functions, but the work of updating all of the analytical solutions in the extensions to the two functions needs to be done as well.

src/sparse.jl
ChrisRackauckas: It's mainly for autodiff.

src/abstract.jl
ChrisRackauckas: I think a mode trait would be much more desirable, indeed since something like Enzyme can have two modes and thus you'd need to split it to two types if you're to do it like this.

gdalle: Yeah it's rather uncomfortable to have it be a subtype of neither forward nor reverse mode. I'll update the PR with a trait mechanism instead, but it's gonna be more breaking as a result

src/sparse.jl
gdalle: okay then it makes sense to put it here

gdalle:
> I like the idea of two functions, but the work of updating all of the analytical solutions in the extensions to the two functions needs to be done as well.

I agree. Maybe ArrayInterface can keep defining `matrix_colors` (based on `ADTypes.column_coloring`), and that way as a transition we can just copy-paste the following lines in the downstream packages:
```julia
ADTypes.column_coloring(M) = ArrayInterface.matrix_colors(M)
ADTypes.row_coloring(M) = ArrayInterface.matrix_colors(M')
```

gdalle:
Thanks for the comments! I'll update the PR with a trait mechanism and ping you again

ChrisRackauckas:
> I agree. Maybe ArrayInterface can keep defining matrix_colors (based on ADTypes.column_coloring), and that way as a transition we can just copy-paste the following lines in the downstream packages:

Let's do the clean break. Such a clean break is the kind of thing that needs someone motivated in order to say yes. I don't see you stopping the DI push, so let's do it right. v0.3.0 here to signal the break, update the downstream in ArrayInterface and SparseDiffTools, and get sparse DI to do things the best way possible, and then integrate it into everything like NonlinearSolve.jl and DifferentialEquations.jl. You've got gusto and we've been sitting here waiting for something like DI to finally fix some of these remaining issues, so let's just pull the trigger and do it.

gdalle:
Not ready yet but close

gdalle:
@ChrisRackauckas I have implemented the trait mechanism and used the opportunity for other small breaking changes in the concrete structs. All of it is summarized in the first comment, which I have edited: https://github.com/SciML/ADTypes.jl/pull/40#issue-2239188179
What do you think?

gdalle:
@avik-pal if you have an opinion I'm curious too

src/dense.jl
```diff
@@ -0,0 +1,243 @@
...
+mode(::AutoFiniteDiff) = FiniteDifferencesMode()
+
+"""
+    AutoFiniteDifferences{T}
+
+Chooses [FiniteDifferences.jl](https://github.com/JuliaDiff/FiniteDifferences.jl).
+
+# Fields
+
+- `fdm::T`: a [`FiniteDifferenceMethod`](https://juliadiff.org/FiniteDifferences.jl/stable/pages/api/#FiniteDifferences.FiniteDifferenceMethod), constructed for instance with `FiniteDifferences.central_fdm`.
```
ChrisRackauckas, line 97: always central?

src/dense.jl
```diff
@@ -0,0 +1,243 @@
...
+
+mode(::AutoReverseDiff) = ReverseMode()
+
+"""
+    AutoTapir
+
+Chooses [Tapir.jl](https://github.com/withbayes/Tapir.jl).
+
+!!! danger
+    This package is experimental, use at your own risk.
```
ChrisRackauckas, line 209: We should let Tapir say that, otherwise this will linger and potentially not be updated for years 😅

src/mode.jl
```diff
@@ -0,0 +1,59 @@
...
+Return the differentiation mode of `ad`, as a subtype of [`AbstractMode`](@ref).
+"""
+function mode end
+
+"""
+    FiniteDifferencesMode
+
+Trait for AD choices that rely on [finite differences](https://en.wikipedia.org/wiki/Finite_difference).
+"""
+struct FiniteDifferencesMode <: AbstractMode end
```
ChrisRackauckas, line 28: Why is this not just forward mode? Finite Differences is just jvps so it's structurally the same

src/sparse.jl
```diff
@@ -0,0 +1,164 @@
+## Sparsity detector
+
+"""
+    AbstractSparsityDetector
+
+Abstract supertype for sparsity pattern detectors, defined for instance in [Symbolics.jl](https://github.com/JuliaSymbolics/Symbolics.jl).
```
ChrisRackauckas, line 6:
```suggestion
Abstract supertype for sparsity pattern detectors
```

src/sparse.jl
```diff
@@ -0,0 +1,164 @@
...
+
+# See also
+
+- [`AbstractSparsityDetector`](@ref)
+"""
+struct NoSparsityDetector <: AbstractSparsityDetector end
+
+jacobian_sparsity(f, x, ::NoSparsityDetector) = ones(Bool, length(f(x)), length(x))
+jacobian_sparsity(f!, y, x, ::NoSparsityDetector) = ones(Bool, length(y), length(x))
+hessian_sparsity(f, x, ::NoSparsityDetector) = ones(Bool, length(x), length(x))
```
ChrisRackauckas, line 43:
```suggestion
jacobian_sparsity(f, x, ::NoSparsityDetector) = trues(Bool, length(f(x)), length(x))
jacobian_sparsity(f!, y, x, ::NoSparsityDetector) = trues(Bool, length(y), length(x))
hessian_sparsity(f, x, ::NoSparsityDetector) = trues(Bool, length(x), length(x))
```

Should it not assume Float64 and instead just be structural?

Project.toml
```diff
@@ -1,13 +1,31 @@
 name = "ADTypes"
 uuid = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
-authors = ["Vaibhav Dixit <vaibhavyashdixit@gmail.com> and contributors"]
-version = "0.2.7"
+authors = [
+    "Vaibhav Dixit <vaibhavyashdixit@gmail.com>, Guillaume Dalle and contributors",
+]
+version = "0.3.0"
```
ChrisRackauckas, line 6:
```suggestion
version = "1.0.0"
```

src/dense.jl
```diff
@@ -0,0 +1,243 @@
...
+# Fields
+
+- `fdm::T`: a [`FiniteDifferenceMethod`](https://juliadiff.org/FiniteDifferences.jl/stable/pages/api/#FiniteDifferences.FiniteDifferenceMethod), constructed for instance with `FiniteDifferences.central_fdm`. 
+
+# Constructor
+
+    AutoFiniteDifferences(; fdm)
+"""
+Base.@kwdef struct AutoFiniteDifferences{T} <: AbstractADType
+    fdm::T
```
Vaibhavdixit02, line 104: I think you might have missed to add a default here?

src/dense.jl
```diff
@@ -0,0 +1,243 @@
...
+
+- `obj_sparse::Bool`
+- `cons_sparse::Bool`
+
+# Constructor
+
+    AutoModelingToolkit(; obj_sparse=false, cons_sparse=false)
+"""
+Base.@kwdef struct AutoModelingToolkit <: AbstractADType
+    obj_sparse::Bool = false
```
Vaibhavdixit02, line 151: This was based on Optimization.jl's use case, if you plan to have a more general implementation we can change the fields

Vaibhavdixit02:
>Should we document type parameters of the concrete structs? I feel like this limits us in the range of non-breaking changes we can do: for instance, turning AutoPolyesterForwardDiff{chunksize} into AutoPolyesterForwardDiff{chunksize,T} is breaking because someone might have used the former as a constructor. That is why I explicitly document keyword constructors and fields.

If we provide constructors for the previous version it can be done in a non-breaking way so it shouldn't be an issue I think

>Is the weak dependency handling okay on Julia < 1.9? The alternative would be to avoid defining the mode trait altogether, which might be preferable

I think that's fine, handling <1.9 on CI and dep versions has been quite a pain already so I think we can bump julia compat 1.9 and do it the "right" way here

>Should we pull in Diffractor.jl and Zygote.jl in the test suite to check on actual instances of the RuleConfig paradigm? I define very simple RuleConfigs in runtests.jl but it's less reliable than taking the real ones

Definitely, though I am not I understand why you are hesitant to do it so maybe I am missing something?

Project.toml
Vaibhavdixit02: Not sure if this is needed right away, can you explain why you'd prefer to skip to 1 directly?

ChrisRackauckas:
One step that we have not done yet but really need to consider in the API is structure identification. While it does not matter too much for the AD aspects, for the real usage of the sparse matrices it can be greatly beneficial to specialize away from SparseMatrixCSC to BandedMatrices or BlockBandedMatrices. So we really should have a step whereby if someone does sparsity detection, it should have a choice of whether to run a structure detection. There is a trade-off here since it's fundamentally dynamic and thus would have a dynamic type choice, but is probably something that is beneficial in "most" uses of sparsity detection. It would need hyperparameters though, for example maximum band size (since all matrices are "banded" if it's large enough), and the algorithm does not exist right now. But it would probably make sense to make some space for this to be a post-processing to sparsity detection

Project.toml
ChrisRackauckas: Pre 1.0 versioning has issues so if we're doing a breaking change then we should go to 1.0

src/dense.jl
gdalle: No, hence the "for instance". Should I remove the hint?

src/dense.jl
gdalle: Fair enough ^^

src/mode.jl
gdalle: Because ADTypes used to have four modes. But I agree that finite differences and forward are not different (they are handled the same by DI). Still I think it makes sense to split, and downstream users can reunite them if they want, as I did

src/sparse.jl
gdalle: I agree with the change but I don't understand the remark, ones(Bool) gives you the same as trues and has nothing to do with floats 🤔 ?

Project.toml
gdalle: I think this design is sound so I would agree. And nothing stops us from going beyond v1 in the future, even if many packages are reluctant to do so

src/mode.jl
ChrisRackauckas: I don't understand what it's supposed to mean then. How would it even be defined? If you make this a part of forward, then it's jvp vs vjp vs element-wise, which has a clear definition. But this is a piece that is jvp, so should it be a subtype of forward mode?

src/dense.jl
gdalle: See the comment at the top, I voluntarily removed it because picking "nothing" doesn't make much sense here. Like for AutoChainRules, I feel like the field of AutoFiniteDifferences has to be a specific type anyway

src/dense.jl
gdalle: I was wondering about that. Shouldn't this be AutoSymbolics? This "obj" and "cons" doesn't make much sense to me in a generic AD setting. Can you suggest an AutoSymbolics struct to replace it?

src/sparse.jl
ChrisRackauckas: I missed the Bool, though why use the extra space? A bitmatrix would be a more memory efficient form of the same structure and since it's not to be computed on and instead just be a structural representation I think it doesn't the accelerated read/write at the cost of size. Its main use would be conversion.

gdalle:
> If we provide constructors for the previous version it can be done in a non-breaking way so it shouldn't be an issue I think

At this point we're breaking so much that I think people who weren't using the kwdef constructors will be able to adapt rather easily. Otherwise we'd have to break type parameter symmetry and define things like 'AutoPolyesterForwardDiff{chunksize}(tag)', but that's not absurd either.

> Definitely, though I am not I understand why you are hesitant to do it so maybe I am missing something?

Just the thought of a heavier test suite, and Diffractor being frequently broken.

> One step that we have not done yet but really need to consider in the API is structure identification.

I think I'd rather keep the functions for sparsity detection simple, and encapsulate this "structure detection vs rely on type alone" in the chosen SparsityDetector object

src/dense.jl
Vaibhavdixit02: Yeah `AutoSymbolics` would be more accurate. With `AutoSparse` it should not need any fields

Vaibhavdixit02: Since it's multiple lines I couldn't suggest a change directly

Vaibhavdixit02:
>Just the thought of a heavier test suite, and Diffractor being frequently broken.

Yeah the second point can get frustrating. Maybe we can keep this as it is for now and if it brings up unforeseen issues we'll add them to really test them here

src/mode.jl
gdalle: Again, I stuck to the original ADTypes design, but if you think that the distinction makes no sense (and I agree) then I'll remove it

ChrisRackauckas: I'd remove it.

src/dense.jl
ChrisRackauckas: I think the hint is confusing.

ChrisRackauckas:
> I think I'd rather keep the functions for sparsity detection simple, and encapsulate this "structure detection vs rely on type alone" in the chosen SparsityDetector object

As long as there is a plan, since I assume DI will need to incorporate it about 1 year from now.

src/sparse.jl
gdalle: good point, I didn't remember that `trues` outputs a `BitMatrix`, I'll make the switch

src/dense.jl
gdalle: removing it

src/dense.jl
gdalle: doing it now

src/dense.jl
gdalle: removing the warning

src/mode.jl
gdalle: okay

gdalle:
For testing, while it's a hassle to add Diffractor and Zygote to the dependencies, I did perform the following check of my `mode` method definitions in the ChainRulesCore extension, so I'm reasonably confident:

```julia
julia> import Diffractor, Zygote

julia> using ADTypes

julia> ADTypes.mode(AutoChainRules(; ruleconfig=Zygote.ZygoteRuleConfig()))
ADTypes.ReverseMode()

julia> ADTypes.mode(AutoChainRules(; ruleconfig=Diffractor.DiffractorRuleConfig()))
ADTypes.ForwardOrReverseMode()

julia> @which ADTypes.mode(AutoChainRules(; ruleconfig=Zygote.ZygoteRuleConfig()))
mode(::AutoChainRules{RC}) where RC<:(ChainRulesCore.RuleConfig{>:ChainRulesCore.HasReverseMode})
     @ ADTypesChainRulesCoreExt ~/.julia/packages/ADTypes/bOklB/ext/ADTypesChainRulesCoreExt.jl:16

julia> @which ADTypes.mode(AutoChainRules(; ruleconfig=Diffractor.DiffractorRuleConfig()))
mode(::AutoChainRules{RC}) where RC<:(ChainRulesCore.RuleConfig{>:Union{ChainRulesCore.HasForwardsMode, ChainRulesCore.HasReverseMode}})
     @ ADTypesChainRulesCoreExt ~/.julia/packages/ADTypes/bOklB/ext/ADTypesChainRulesCoreExt.jl:22
```

Of course there is no ChainRules-compatible backend with a forward-only `RuleConfig`, so we'll never hit the third method

gdalle:
I think we're there, the last few things to iron out are:

- Should we bump Julia compat to 1.9? I don't see the need, especially if large parts of SciML end up relying on ADTypes. It does mean that two unconditional dependencies (EnzymeCore and ChainRulesCore) will be added everywhere due to the lack of extensions on versions <1.9, but I think this beats erroring? Alternately, we could skip the `mode` definition altogether and leave it to downstream users.
- Should we bump ADTypes version to 1.0?
- Should we leave JET in the test suite, even though it errors on nightly? It does a great job of catching typos, and I made it so that the badge would still be green if only nightly fails

ChrisRackauckas:
> If we define the mode, we need extensions for ChainRulesCore and EnzymeCore (fairly lightweight). Should we

We should define mode and at least bump to v1.9. I'd say just go to v1.10 knowing it's the next LTS and skip all of the other pain, there's other stuff to worry about in life. 

> Should we bump ADTypes version to 1.0?

Yes

> Should we leave JET in the test suite, even though it errors on nightly? It does a great job of catching typos, and I made it so that the badge would still be green if only nightly fails

We just shouldn't run nightly, it's not for humans. We can set the new prerelease branch when it's ready, but since nightly isn't for packages (which is why the prerelease is being made) it's a waste of our time to be trying it.

gdalle:
> I'd say just go to v1.10 knowing it's the next LTS and skip all of the other pain, there's other stuff to worry about in life.

Done.

> We just shouldn't run nightly, it's not for humans.

Removed nightly from CI

gdalle:
Once I get an approving review I'll merge, but I'd like to do two things before we register v1:

- ask for feedback, either on Slack or Discourse
- test it with DifferentiationInterface to make sure we didn't forget something essential

ChrisRackauckas (approved):

ChrisRackauckas:
Yes no need to release right away, let's get downstream all set and make sure everyone is bought into this form.

Vaibhavdixit02 (approved):
🎉🎉

Issue #47, opened 2024-04-22
Support older Julia versions?


lassepe:
The main thing that stops me from adopting this package is that I don't want to force all my users to upgrade Julia 1.10. ~~As I understand it, dropping support for older Julia versions would be considered a breaking change.~~ I see no fundamental reason why this package would require 1.10 --- package extensions could be used in a backward compatible way.

gdalle:
This was discussed in https://github.com/SciML/ADTypes.jl/pull/40#issuecomment-2055319884 and @ChrisRackauckas suggested we drop Julia versions below 1.10. On second thought, if this is a real pain point for users, I guess there is not much harm in adding back support all the way down to 1.6

gdalle:
After some more discussion, older versions will not be supported, not with a v1.10 LTS around the corner

lassepe:
Okay, sounds good! Also good to know that 1.10 will be the new LTS :)

PR #48, opened 2024-04-22
Support Julia 1.6 (again)

gdalle:
## Checklist

## Additional context

This PR makes EnzymeCoreExt and ChainRulesExt unconditional dependencies for Julia <1.9, thus allowing support for older Julia versions.
We initially decided against it in https://github.com/SciML/ADTypes.jl/pull/40#issuecomment-2055319884, but as shown by #47 some users would like this feature. It is zero-cost for us, so why not?

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - '1.10'
+          - '1.6'
           - '1'
         os:
           - ubuntu-latest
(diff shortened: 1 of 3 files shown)


gdalle:
@ChrisRackauckas @Vaibhavdixit02 I think this doesn't hurt, right?

gdalle:
Bump @Vaibhavdixit02

ChrisRackauckas:
If you all want to maintain it, go for it. We don't plan to put any effort into the old LTS since the new one is already out.

gdalle:
I didnt have that in mind, so all of SciML is basically willing to drop 1.6?

ChrisRackauckas:
Yes.

PR #50, opened 2024-04-30
ci: standardise workflows using SciML's reusable workflows

thazhemadam:
Update the workflows in this repository to use SciML's reusable workflows.
This is part of a larger effort to standardise the SciML's CI workflows for more generic and common requirements, to keep the workflows uniform and easier to maintain.

### DIFF
--- a/.github/workflows/Documentation.yml
+++ b/.github/workflows/Documentation.yml
@@ -1,4 +1,4 @@
-name: Documentation
+name: "Documentation"
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


gdalle:
Where does this build the docs? And are they updated directly upon each merged PR?

thazhemadam:
The documentation is still built on the same events as it is at present, i.e., on pushes to `main`, on all tags pushed and on all pull requests opened. The status quo is expected to remain unchanged. https://github.com/SciML/ADTypes.jl/blob/0f86da069c526ad1b785aca415026e1610c534ff/.github/workflows/Documentation.yml#L3-L8

gdalle:
I mean on which website, does it go to a centralized SciML docs or does it stay here?

thazhemadam:
Ah, I see. Sorry about the misunderstanding. The documentation deployments will continue to stay here, as specified in the `deploydocs` method.
https://github.com/SciML/ADTypes.jl/blob/a67c600dacccdf1c55b9e975bdc0b4621fec2dd4/docs/make.jl#L11

Vaibhavdixit02:
Even though it passed tests in the PR the CI on master seems to have broken can you take a look @thazhemadam?

gdalle:
I re-ran the failed workflows and it worked, I think it was just a Codecov glitch

PR #55, opened 2024-05-14
Add safe mode

willtebbutt:
## Checklist

## Additional context

Tapir.jl now has a "safe mode", in which lots of additional type checking is performed, in addition to some value checks. I would like to enable it by default (I'll be adding `@info` calls on the Tapir end to make it clear to users that they're using safe mode). This is to ensure that users are aware of this feature, which often makes debugging a lot easier, and can pick up on issues before they propagate to some arbitrary point further on in the code.

Provided that people are happy with this, if someone could advise on the appropriate version bump for this package I would appreciate it.

### DIFF
--- a/test/dense.jl
+++ b/test/dense.jl
@@ -138,6 +138,7 @@ end
     @test ad isa AbstractADType
     @test ad isa AutoTapir
     @test mode(ad) isa ReverseMode
+    @test ad.safe_mode == true
 end
 
 @testset "AutoTracker" begin
--- a/test/dense.jl
+++ b/test/dense.jl
@@ -138,7 +138,10 @@ end
     @test ad isa AbstractADType
     @test ad isa AutoTapir
     @test mode(ad) isa ReverseMode
-    @test ad.safe_mode == true
+    @test ad.safe_mode
+    
+    ad = AutoTapir(; safe_mode=false)
+    @test !ad.safe_mode
 end
 
 @testset "AutoTracker" begin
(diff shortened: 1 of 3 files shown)


gdalle:
Would it make sense for it to be a type parameter with two possible values, instead of a field? In other words, does it bring any benefit to have that information statically available?

willtebbutt:
Good question.

It shouldn't matter for performance, because whether or not a rule operates in safe mode is baked in when you call `build_rrule` (extra code is injected during codegen to check pre- and post-conditions of every call to a rule that happens inside a derived rule on both the forwards and reverse passes -- this is propagated through the entire compiled function so that everything gets checked).

On a practical level, (I think) adding type parameters would _technically_ be breaking, so it would be good to avoid if possible.

gdalle:
Okay then you can bump the minor version and I'll merge this

willtebbutt:
Great, thanks!

test/dense.jl
```diff
@@ -138,6 +138,7 @@ end
     @test ad isa AbstractADType
     @test ad isa AutoTapir
     @test mode(ad) isa ReverseMode
+    @test ad.safe_mode == true
```
gdalle, line 141:
```suggestion
    @test ad.safe_mode
    
    ad = AutoTapir(; safe_mode=false)
    @test !ad.safe_mode
```

gdalle (changes requested):

test/dense.jl
willtebbutt: Good idea.

gdalle (approved):

gdalle:
Tests only fail due to codecov throttling, merging now

PR #60, opened 2024-05-26
CI improvements

ArnoStrouwen:
Makes CI more uniform with rest of SciML.
Removes testing for 1.6, since rest of SciML dropped support.
Activates errors on deprecation. Not sure what to do with the tests in the legacy file, since these will now error.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,7 +20,7 @@ ADTypesEnzymeCoreExt = "EnzymeCore"
 [compat]
 ChainRulesCore = "1.0.2"
 EnzymeCore = "0.5.3,0.6,0.7"
-julia = "1.6"
+julia = "1.10"
 
 [extras]
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -52,13 +52,11 @@ end
 ## Tests
 
 @testset verbose=true "ADTypes.jl" begin
-    if VERSION >= v"1.10"
-        @testset "Aqua.jl" begin
-            Aqua.test_all(ADTypes; deps_compat = (check_extras = false,))
-        end
-        @testset "JET.jl" begin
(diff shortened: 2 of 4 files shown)


.github/workflows/CI.yml
```diff
@@ -1,42 +1,47 @@
...
     runs-on: ${{ matrix.os }}
+    env:
+      GROUP: ${{ matrix.group }}
     strategy:
       fail-fast: false
       matrix:
+        group:
+          - Core
         version:
-          - '1.6'
```
gdalle, line 21: ADTypes is compatible with 1.6 so it needs to be tested on 1.6. There are packages outside of SciML that depend on it and want to support 1.6 (first and foremost DifferentiationInterface). I don't approve this change

gdalle (changes requested):

Vaibhavdixit02:
I think we want to allow the deprecation warnings to pass here. Otherwise, we'd need downstream breaking releases which isn't ideal

gdalle:
Why do you want to activate error on deprecation? We deprecated things on purpose in v1 to allow for a smooth transition 

@ChrisRackauckas this is a disruptive PR, please don't merge until I'm okay with it

Project.toml
```diff
@@ -20,7 +20,7 @@ ADTypesEnzymeCoreExt = "EnzymeCore"
 [compat]
 ChainRulesCore = "1.0.2"
 EnzymeCore = "0.5.3,0.6,0.7"
-julia = "1.6"
```
gdalle, line 23: I need 1.6 support

test/runtests.jl
```diff
@@ -52,13 +52,11 @@ end
 ## Tests
 
 @testset verbose=true "ADTypes.jl" begin
-    if VERSION >= v"1.10"
```
gdalle, line 55: Essential for tests on 1.6 where JET doesn't work

gdalle (changes requested):

PR #61, opened 2024-06-06
Fix constructor of AutoMTK

Vaibhavdixit02:
## Checklist

- [ ] Appropriate tests were added
- [ ] All documentation related to code changes were updated
- [ ] The new code follows the
- [ ] Any new documentation only uses public API
  
## Additional context

I was overloading the constructor in OptimizationBase since the deprecation here wasn't sufficient, leading to re-precompilation. I couldn't find a way to use `@deprecated` for this so this'll have to do.


gdalle:
I'll review tomorrow, please wait for me

src/legacy.jl
```diff
@@ -11,4 +11,11 @@
 
 @deprecate AutoSparseZygote() AutoSparse(AutoZygote())
 
-@deprecate AutoModelingToolkit(; kwargs...) AutoSparse(AutoSymbolics())
+function AutoModelingToolkit(sparse = false, cons_sparse = false; kwargs...)
+    @warn "AutoModelingToolkit(args...) is deprecated, use AutoSymbolics() or AutoSparse(AutoSymbolics()) instead."
```
ChrisRackauckas, line 15:
```suggestion
    @warn "AutoModelingToolkit(args...) is deprecated, use AutoSymbolics() or AutoSparse(AutoSymbolics()) instead."  maxlog=1
```

ChrisRackauckas (approved):
Just needs the maxlog thing and this is good.

Vaibhavdixit02:
Thanks, I wasn't completely sure what you meant by it in the chat

Vaibhavdixit02:
I'll wait for @gdalle to take a look to make sure it doesn't affect DI

test/legacy.jl
```diff
@@ -1,5 +1,7 @@
-@testset "AutoModelingToolkig" begin
-    ad = @test_deprecated AutoModelingToolkit()
+@testset "AutoModelingToolkit" begin
```
gdalle, line 1: add tests for the keyword arguments

gdalle (commented):
I don't see the point of introducing a new object when it's already deprecated, and it doesn't correspond to an actual autodiff backend.
We made a breaking release of ADTypes, a breaking release of OptimizationBase might be coming, we should leverage that instead of trying to minimize breakage by sticking to old patterns.
If you need a shortcut for Symbolics with or without sparsity in OptimizationBase, why not use `AutoModelingToolkitDeprecated` or something of that sort, which you define yourself?

gdalle:
The `AutoModelingToolkit` struct in ADTypes v0.2.7 was as follows, so this is the full API we want to deprecate but with meaningful fallbacks:
https://github.com/SciML/ADTypes.jl/blob/9f17b8227afdb12cbeb0faa7edc18abd5ffcae0c/src/ADTypes.jl#L151-L164
@ChrisRackauckas is right that the deprecation logic wasn't correct, so my last commit fixes it. We still generate a deprecation warning, which will be visible to users if such warnings are enabled, but we fall back on the correct version of the Symbolics backend.
@Vaibhavdixit02 what do you think?

ChrisRackauckas:
Why did you change it back to being incorrect? The point is that there was a version without kwargs before.

gdalle:
My fallbacks handle every version that there was before, with and without kwargs. Both will give the correct result, but both will be deprecated

ChrisRackauckas (approved):

ChrisRackauckas:
I didn't see the other dispatch. This should work then.

gdalle:
Let's wait for Vaibhav to test it locally with OptimizationBase before merging

PR #63, opened 2024-06-22
Promote ReverseDiff compile field to type

avik-pal:
Needed for things like https://github.com/LuxDL/Lux.jl/issues/642 where the return type needs to rely on whether the compiled tape needs to be returned.

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -201,6 +201,15 @@ struct AutoReverseDiff{C} <: AbstractADType
     end
 end
 
+function Base.getproperty(ad::AutoReverseDiff, s::Symbol)
+    if s === :compile
+        Base.depwarn(
+            "`ad.compile` where `ad` is `AutoReverseDiff` has been deprecated and will be removed in v2. Instead it is available as a compile-time constant as `AutoReverseDiff{true}` or `AutoReverseDiff{false}`.",
+            :getproperty)
+    end
+    return getfield(ad, s)
+end
+
 mode(::AutoReverseDiff) = ReverseMode()
 
 """
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -186,14 +186,19 @@ Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
 
 # Constructors
 
-    AutoReverseDiff(; compile=false)
+    AutoReverseDiff(; compile::Union{Val, Bool} = Val(false))
 
 # Fields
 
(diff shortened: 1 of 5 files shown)


Vaibhavdixit02 (approved):

src/dense.jl
```diff
@@ -186,14 +186,19 @@ Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
 
 # Constructors
 
-    AutoReverseDiff(; compile=false)
+    AutoReverseDiff(; compile::Union{Val, Bool} = Val(false))
```
gdalle, line 189: changing the default value of `compile` from `false` to `Val(false)` is breaking, we should keep `false`

gdalle (changes requested):

src/dense.jl
ChrisRackauckas: No it's not because it also handles the bool.

ChrisRackauckas (approved):
This is s a nice non-breaking way to handle it. And the bool route with constant prop still achieves the same effect.

ChrisRackauckas:
One change, getproperty overload to add a depwarn on accessing the boolean. We'll want people to move to the type stable form, though there's no reason to force the change immediately (and that depwarn should probably be given at least a year or two given that the feature has been around since 2020)

src/dense.jl
avik-pal: That wouldn't be breaking because the final object being created is what is accessible to the user, and there is no way to distinguish between `AutoReverseDiff(; compile=false)` and `AutoReverseDiff(; compile=Val(false))` (except the case I mention next).

Let's say we keep `compile = false`, then the following fails:

```julia
@inferred AutoReverseDiff()
```

But changing default to `Val` makes it indistinguishable.

gdalle: There is a way to distinguish these objects, which is when you access the compile field and expect a Bool. That is part of the current API and users of ADTypes (like DI) rely on it.

gdalle:
Yes it is breaking, see https://github.com/SciML/ADTypes.jl/pull/63#discussion_r1649897718

src/dense.jl
avik-pal: it is a bool

avik-pal:
> Yes it is breaking, see https://github.com/SciML/ADTypes.jl/pull/63#discussion_r1649897718

How though? ad.compile is a Bool as it always was.

gdalle:
Oh sorry I missed that part, thought you stored the Val directly. Then it's fine

src/dense.jl
gdalle: Right, my bad

PR #70, opened 2024-07-13
Document new type parameter for AutoReverseDiff

gdalle:
Complete #63 by documenting the type parameter, making it part of the API


ChrisRackauckas (approved):
Yeah I'd be happy to extend the API for this. @avik-pal agreement?

gdalle:
If it's not in the API it's an internal and no other package can rely on it, so there would be no point in introducing it in the first place.

avik-pal (approved):

PR #72, opened 2024-07-17
Add constant_function kwarg to AutoEnzyme

gdalle:
## Checklist

## Additional context

This PR adds a keyword argument (and type parameter) to `AutoEnzyme`, which specified whether the function itself should be considered constant during differentiation. See <https://github.com/SciML/SciMLBenchmarks.jl/pull/988#issuecomment-2228969526> for the related discussion with @wsmoses and @ChrisRackauckas

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -50,7 +50,7 @@ Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
     AutoEnzyme(; mode=nothing, constant_function::Bool=false)
 
 The `constant_function` keyword argument (and type parameter) determines whether the function object itself should be considered constant or not during differentiation with Enzyme.jl.
-For simple functions, this should usually be set to `false`, but in the case of closures or callable structs which contain differentiated data, it should be set to `true`.
+For simple functions, `constant_function` should usually be set to `false`, but in the case of closures or callable structs which contain differentiated data that can be treated as constant, `constant_function` should be set to `true` for increased performance.
 
 # Fields
 
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -39,7 +39,7 @@ struct AutoDiffractor <: AbstractADType end
 mode(::AutoDiffractor) = ForwardOrReverseMode()
 
 """
-    AutoEnzyme{M}
+    AutoEnzyme{M,constant_function}
 
 Struct used to select the [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl) backend for automatic differentiation.
 
@@ -47,7 +47,10 @@ Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
 
 # Constructors
 
-    AutoEnzyme(; mode=nothing)
+    AutoEnzyme(; mode=nothing, constant_function::Bool=false)
+
+The `constant_function` keyword argument (and type parameter) determines whether the function object itself should be considered constant or not during differentiation with Enzyme.jl.
+For simple functions, this should usually be set to `false`, but in the case of closures or callable structs which contain differentiated data, it should be set to `true`.
 
 # Fields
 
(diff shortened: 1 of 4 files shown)


src/dense.jl
```diff
@@ -39,15 +39,18 @@ struct AutoDiffractor <: AbstractADType end
...
 
 Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
 
 # Constructors
 
-    AutoEnzyme(; mode=nothing)
+    AutoEnzyme(; mode=nothing, constant_function::Bool=false)
+
+The `constant_function` keyword argument (and type parameter) determines whether the function object itself should be considered constant or not during differentiation with Enzyme.jl.
+For simple functions, this should usually be set to `false`, but in the case of closures or callable structs which contain differentiated data, it should be set to `true`.
```
ChrisRackauckas, line 53:
```suggestion
For simple functions, this should usually be set to `false`, but in the case of closures or callable structs which contain differentiated data should be treated as constant, it should be set to `true`.
```

src/dense.jl
```diff
@@ -39,15 +39,18 @@ struct AutoDiffractor <: AbstractADType end
...
 
 Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
 
 # Constructors
 
-    AutoEnzyme(; mode=nothing)
+    AutoEnzyme(; mode=nothing, constant_function::Bool=false)
+
+The `constant_function` keyword argument (and type parameter) determines whether the function object itself should be considered constant or not during differentiation with Enzyme.jl.
+For simple functions, `constant_function` should usually be set to `false`, but in the case of closures or callable structs which contain differentiated data that can be treated as constant, `constant_function` should be set to `true` for increased performance.
```
ChrisRackauckas, line 53:
```suggestion
For simple functions, `constant_function` should usually be set to `false`, but in the case of closures or callable structs which contain differentiated data that can be treated as constant, `constant_function` should be set to `true` for increased performance.

!!! note
    If `constant_function` is set to true but the enclosed data is not truly constant, then Enzyme will not compute the
    correct derivative values. An example of such a function is:
    
    ```julia
    cache = [0.0]
    function f(x)
       cache[1] = x[1]^2
       cache[1] + x[1]
    end
    ```

    In this case, the enclosed cache is a function of the differentiated input, and thus its values are non-constant with
    respect to the input. Thus in order to compute the correct derivative of the output, the derivative must propagate
    through the `cache` value and it must not be treated as constant. Alternatively, the following function can treat
    `parameter` as a constant:

    ```julia
    parameter = [0.0]
    function f(x)
       parameter[1] + x[1]
    end
    ```

    because `parameter` is never modified based on the input `x`. In this case, `constant_function = true` would allow
    the chosen differentiation system to perform extra memory and compute optimizations under the assumption
    `parameter` is kept constant. 

```

ChrisRackauckas (approved):

src/dense.jl
```diff
@@ -39,25 +39,63 @@ struct AutoDiffractor <: AbstractADType end
...
 
 Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
 
 # Constructors
 
-    AutoEnzyme(; mode=nothing)
+    AutoEnzyme(; mode=nothing, constant_function::Bool=false)
+
+The `constant_function` keyword argument (and type parameter) determines whether the function object itself should be considered constant or not during differentiation with Enzyme.jl.
+For simple functions, `constant_function` should usually be set to `false`, but in the case of closures or callable structs which contain differentiated data that can be treated as constant, `constant_function` should be set to `true` for increased performance (more details below).
```
wsmoses, line 53: This should read be usually set to true. However if the closure has differentiable data loaded and stored into, it may need to be set false

gdalle: Woops I switched true and false

src/dense.jl
```diff
@@ -39,25 +39,63 @@ struct AutoDiffractor <: AbstractADType end
...
 # Fields
 
   - `mode::M`: can be either
 
       + an object subtyping `EnzymeCore.Mode` (like `EnzymeCore.Forward` or `EnzymeCore.Reverse`) if a specific mode is required
       + `nothing` to choose the best mode automatically
+
+# Notes
+
+If `constant_function = true` but the enclosed data is not truly constant, then Enzyme.jl will not compute the correct derivative values.
```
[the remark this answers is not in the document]
gdalle: I understand that you always want to stress that Enzyme computes the right thing, but this is a high-level docstring for beginners. I don't think we're implying that Enzyme is wrong here, just that someone might be using it wrong if they say `constant_function=true`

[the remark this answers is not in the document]
gdalle: I tried to be clearer in #73

PR #73, opened 2024-07-17
Fix AutoEnzyme docstring for constant_function

gdalle:
There was a typo in #72 for the docstring, I switched `true` and `false` in the explanation of `constant_function`

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -50,7 +50,8 @@ Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
     AutoEnzyme(; mode=nothing, constant_function::Bool=false)
 
 The `constant_function` keyword argument (and type parameter) determines whether the function object itself should be considered constant or not during differentiation with Enzyme.jl.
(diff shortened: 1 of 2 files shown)


gdalle:
@willtebbutt I wonder if the same kind of distinction would make sense for Tapir? To clarify, we're introducing this for Enzyme so that DifferentiationInterface can know whether to `Duplicate`/shadow the function `f` itself or not.

src/dense.jl
```diff
@@ -61,30 +62,39 @@ For simple functions, `constant_function` should usually be set to `false`, but
...
+    return x[1]
 end
 ```
 
-In this case, the enclosed cache is a function of the differentiated input, and thus its values are non-constant with respect to the input.
-Thus, in order to compute the correct derivative of the output, the derivative must propagate through the `cache` value, and said `cache` must not be treated as constant.
-
-Conversely, the following function can treat `parameter` as a constant, because `parameter` is never modified based on the input `x`:
+The function `f1` is not a closure, it does not contain any data.
+Thus `f1` can be differentiated with `AutoEnzyme(constant_function=true)`.
```
ChrisRackauckas, line 75:
```suggestion
Thus `constant_function` does not change the behavior or performance of this function.
```

ChrisRackauckas (approved):

PR #74, opened 2024-08-02
Remove `constant_function` for `AutoEnzyme`

gdalle:
> [!WARNING]
> This PR and the associated release are breaking wrt 1.6.1, but we will yank 1.6.1 and 1.6.0 right after

## Checklist

## Additional context

Following a long discussion with @ChrisRackauckas and @wsmoses, we decided to remove the `constant_function` setting for now, and reinsert it later with different defaults (once Enzyme implements a few changes on their side). This PR does the removal and bumps ADTypes to v1.6.2, after which we will need to yank v1.6.0 and v1.6.1.

- https://github.com/EnzymeAD/Enzyme.jl/issues/1669

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
 authors = [
     "Vaibhav Dixit <vaibhavyashdixit@gmail.com>, Guillaume Dalle and contributors",
 ]
-version = "1.6.1"
+version = "1.6.2"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -39,7 +39,7 @@ struct AutoDiffractor <: AbstractADType end
 mode(::AutoDiffractor) = ForwardOrReverseMode()
 
 """
-    AutoEnzyme{M,constant_function}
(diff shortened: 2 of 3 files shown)


gdalle:
Let's wait for a review from Billy and Chris before merging

ChrisRackauckas:
Let's just go straight to the `nothing` default then?

gdalle:
I wanted to wait until the implementation exists on the Enzyme and DI side before reintroducing this kwarg. Our mistake last time was to introduce it before it actually did anything, and as a result we got the design wrong.

gdalle:
But if @wsmoses is confident that `constant_function=nothing` is the right third option I can modify this PR to just change the default instead of removing the kwarg

wsmoses:
I would probably do the yank now then design when ready.

Being even more forward looking this should likely actually be a Union{Nothing, Enzyme.Annotation}, since the function could be nothing, constant, duplicated, or even active

wsmoses:
This constant function is not the appropriate name

ChrisRackauckas:
function_activity

wsmoses:
I mean that sounds like a good name to me, but it also might be wise to
wait til things are settled before choosing

gdalle:
Okay then removing everything first is the right move. In this spirit, do you approve this PR?

wsmoses (approved):
Spiritually and actually approved

PR #75, opened 2024-08-02
Deprecate "safe_mode" in favour of "debug_mode" for Tapir.jl

willtebbutt:
## Checklist

## Additional context

Also deprecates a default choice in favour of requiring the user to specific whether they want `debug_mode` on / off.

The implementation is quite ugly, because we cannot use the usual `@deprecate` macro because Julia does not specialise on kwargs -- if you attempt to use the `@deprecate` macro, you'll find that you wind up overwriting existing methods, precompilation fails, and everything breaks. The only way I've been able to find to make this work is to create the new method with two kwargs added in this PR, in which each case is handled explicitly.

Unfortunately we cannot rename the `safe_mode` field of the `AutoTapir` type to `debug_mode` -- that will have to wait until 2.0. As a result I'm really not 100% convinced by this PR, but I'm going to let @gdalle and @yebai debate its merits.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "47edcb42-4c32-4615-8424-f2b9edc5f35b"
 authors = [
     "Vaibhav Dixit <vaibhavyashdixit@gmail.com>, Guillaume Dalle and contributors",
 ]
-version = "1.6.1"
+version = "1.6.2"
 
 [deps]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -331,11 +331,12 @@ Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
 
 # Constructors
 
-    AutoTapir(; safe_mode=true)
+    AutoTapir(; debug_mode::Bool)
 
 # Fields
 
-  - `safe_mode::Bool`: whether to run additional checks to catch errors early. While this is
+  - `safe_mode::Bool`: (to be renamed to `debug_mode` in the next breaking release)
+    whether to run additional checks to catch errors early. While this is
     on by default to ensure that users are aware of this option, you should generally turn
     it off for actual use, as it has substantial performance implications.
     If you encounter a problem with using Tapir (it fails to differentiate a function, or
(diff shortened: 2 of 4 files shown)


gdalle:
I don't think a v2.0 of ADTypes is anywhere near on the horizon. The switch to v1.0 was only mildly breaking and it was already a nightmare (I should know, I helped handle most of it). I'm making an executive call here and saying that it will remain called "safe mode" in ADTypes for the foreseeable future. The translation between that and whatever Tapir does will happen in DifferentiationInterface, so feel free to make changes there or to update the docstring here. I'm closing this PR however.

gdalle:
I mean, that's what package admins do, they approve changes or they don't. I'm a package admin of ADTypes and I don't approve this specific change, because it's too much hassle for too little gain (even the PR author was not convinced).

Breaking changes can definitely happen in AD packages, and they regularly do. However, our goal is to avoid a breaking release of ADTypes, because that would be extremely disruptive to the whole ecosystem (the last one took weeks to apply and a lot of sweat). If you are an AD developer and want your package to be present in ADTypes (and DifferentiationInterface), that means you need to keep working with some design decisions you made a while ago, even if they are not ideal in retrospect.
In this particular instance, renaming a keyword argument because you named it wrong the first time is not important enough to warrant a breaking release. Therefore it's up to you to do the translation inside Tapir (or DifferentiationInterface's Tapir extension, whichever you prefer).

gdalle:
As for the types "belonging to the individual packages", it's like saying that every package should maintain the DifferentiationInterface bindings. It's great in theory, but in practice you need a centralized push (in this case from me) to make things happen. Maybe in a distant future where DI is stable and widely adopted we can decentralize, but definitely not now.

ChrisRackauckas:
We can still deprecate without breaking. Since Tapir is generally not used doing a break on its semantics would not be hard to update the world to, we could just blind merge almost all PRs.

ChrisRackauckas:
It's not a good thing that's for sure. But Tapir isn't in production mode, so I'd be willing to... let it slide... Might as well change the field name too and set a getproperty overload for that deprecation. Or let @willtebbutt make the call if we can just do a break to Tapir users. Since the set of users is currently probably no more than a handful which includes us in the thread and mostly for benchmarks, it falls into the "experimental" territory and I'd let the decision be up to @willtebbutt, since I'd hate to have this 3 years from now because we didn't want to break 1 benchmark code.

willtebbutt:
Personally, I would rather either
1. do nothing, or
2. make a breaking change to the Tapir implementation.
but I'd really rather avoid this middle ground where we have a lot of ugly code that deprecates stuff for questionable gains.

I'm also keen to change the behaviour to _require_ users to specific `debug_mode`, rather than giving it a default value, which we can't do without a breaking change.

So if we're actually happy to make a breaking change and not bumping the major version of ADTypes.jl, on the basis that Tapir.jl support is "experimental", then I'm in favour of going ahead with that.

ChrisRackauckas:
Let's do it then.

PR #77, opened 2024-08-08
Add `function_annotation` to `AutoEnzyme`

gdalle:
## Checklist

## Additional context

This PR adds a type parameter `A` to `AutoEnzyme` and a keyword argument `function_annotation::Type{A}` to its constructor.
It is the second attempt at properly solving the following issues:

- https://github.com/EnzymeAD/Enzyme.jl/issues/1669
- https://github.com/gdalle/DifferentiationInterface.jl/issues/339

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -39,7 +39,7 @@ struct AutoDiffractor <: AbstractADType end
 mode(::AutoDiffractor) = ForwardOrReverseMode()
 
 """
-    AutoEnzyme{M}
+    AutoEnzyme{M,A}
 
 Struct used to select the [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl) backend for automatic differentiation.
 
@@ -47,28 +47,37 @@ Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
 
 # Constructors
 
-    AutoEnzyme(; mode=nothing)
+    AutoEnzyme(; mode::M=nothing, function_annotation::Type{A}=Nothing)
+
+# Type parameters
+
+  - `A` determines how the function `f` to differentiate is passed to Enzyme. It can be:
+
+      + a subtype of `EnzymeCore.Annotation` (like `EnzymeCore.Const` or `EnzymeCore.Duplicated`) to enforce a given annotation
+      + `Nothing` to simply pass `f` and let Enzyme choose the most appropriate annotation
 
 # Fields
 
-  - `mode::M`: can be either
+  - `mode::M` determines the autodiff mode (forward or reverse). It can be:
 
       + an object subtyping `EnzymeCore.Mode` (like `EnzymeCore.Forward` or `EnzymeCore.Reverse`) if a specific mode is required
       + `nothing` to choose the best mode automatically
 """
-struct AutoEnzyme{M} <: AbstractADType
+struct AutoEnzyme{M, A} <: AbstractADType
     mode::M
 end
 
-function AutoEnzyme(; mode::M = nothing) where {M}
-    return AutoEnzyme{M}(mode)
+function AutoEnzyme(;
+        mode::M = nothing, function_annotation::Type{A} = Nothing) where {M, A}
+    return AutoEnzyme{M, A}(mode)
 end
 
 mode(::AutoEnzyme) = ForwardOrReverseMode()  # specialized in the extension
 
-function Base.show(io::IO, backend::AutoEnzyme)
+function Base.show(io::IO, backend::AutoEnzyme{M, A}) where {M, A}
     print(io, AutoEnzyme, "(")
-    !isnothing(backend.mode) && print(io, "mode=", repr(backend.mode; context = io))
+    !isnothing(backend.mode) && print(io, "mode=", repr(backend.mode; context = io), ", ")
+    print(io, "function_annotation=", repr(A; context = io))
     print(io, ")")
 end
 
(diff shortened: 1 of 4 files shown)


src/dense.jl
```diff
@@ -39,36 +39,46 @@ struct AutoDiffractor <: AbstractADType end
...
 """
-struct AutoEnzyme{M} <: AbstractADType
+struct AutoEnzyme{M, A} <: AbstractADType
     mode::M
 end
 
-function AutoEnzyme(; mode::M = nothing) where {M}
-    return AutoEnzyme{M}(mode)
+function AutoEnzyme(;
+        mode::M = nothing, function_annotation::Type{A} = Nothing) where {M, A}
```
wsmoses, line 71: let's make this more strict to start with, something like

```
A<:Union{Nothing, EnzymeCore.Annotation}
```

test/misc.jl
```diff
@@ -27,6 +27,8 @@ for backend in [
     ADTypes.AutoDiffractor(),
     ADTypes.AutoEnzyme(),
     ADTypes.AutoEnzyme(mode = :forward),
+    ADTypes.AutoEnzyme(function_annotation = Val{:forward}),
+    ADTypes.AutoEnzyme(mode = :reverse, function_annotation = Val{:duplicated}),
```
wsmoses, line 31: this isn't supported by the spec above (which requires annotation, not a symbol), no?

src/dense.jl
gdalle: I can't do this because it would require pulling in EnzymeCore as a dependency of ADTypes, which we want to keep as light-weight and backend-independent as possible

test/misc.jl
gdalle: Yeah this is just for testing printing, I can switch

src/dense.jl
wsmoses: could you make the constructor here just always be the one arg constructor, and move the two arg one (with type checking) to the enzymeext?

wsmoses: actually fwiw, this might also apply to mode. Maybe just keep this as a 0-arg constructor, and move the kwarg version to the ext?

gdalle: That would be breaking because users may want to use the kwarg constructor with nothing, which doesn't require having the extension loaded

gdalle: I agree than in most cases people who call AutoEnzyme will have EnzymeCore loaded, but it's not a requirement

wsmoses: ah right, darn. We should at least try to add a type check to the new arg tho

gdalle: I don't see how to do it without EnzymeCore?

gdalle: I can do it in DI though, since that's mostly how people will use this

wsmoses: yeah I think this is a question of what you can do to put as much of the new arg in an extension

test/misc.jl
gdalle: Actually I'd like to keep it this way for easy visual check. The annotation from EnzymeCore doesn't change much but if I also provide a realistic `mode` argument the print can get rather long and it's less obvious whether there's e.g. a comma mistake

wsmoses: I get that, but on the other hand I'm worried that someone will see this, and assume that this is valid/the way it's supposed to be specified. And per the docstring, this is explicitly not supported.

gdalle: Exactly, so why would they look at the test and not trust the docstring?

gdalle: I can add a comment above the test saying "DONT DO THIS, THIS IS BAD, LOOK AD THE DOCSTRING FOR PROPER KWARG DOCTRINE."

wsmoses: I guess I could be okay with such a comment, but isn't just using it properly easier?

gdalle: Then we'd have to use it properly for every other backend. Bring in FiniteDifferences to have a proper fdm object for AutoFiniteDifferences, give a proper RuleConfig to AutoChainRules, etc.
This is just for visual checking of the display code and it actually helps to keep things concise for our eyes to assess the result, so I really don't think we need to worry that much.

wsmoses: Okay, then I'm fine with this [but a comment warning users would be appreciated -- since I know a bunch of folks that just copy code from tests and try to use]

gdalle: Right now the printout from the test is the following, each backend fits nicely on one line and we see what's missing without scrolling like a madman:

```julia
AutoChainRules(ruleconfig=:rc)
AutoDiffractor()
AutoEnzyme()
AutoEnzyme(mode=:forward)
AutoEnzyme(function_annotation=Val{:forward})
AutoEnzyme(mode=:reverse, function_annotation=Val{:duplicated})
AutoFastDifferentiation()
AutoFiniteDiff()
AutoFiniteDiff(fdtype=:fd, fdjtype=:fdj, fdhtype=:fdh)
AutoFiniteDifferences(fdm=:fdm)
AutoForwardDiff()
AutoForwardDiff(chunksize=3, tag=:tag)
AutoPolyesterForwardDiff()
AutoPolyesterForwardDiff(chunksize=3, tag=:tag)
AutoReverseDiff()
AutoReverseDiff(compile=true)
AutoSymbolics()
AutoTapir()
AutoTapir(safe_mode=false)
AutoTracker()
AutoZygote()
AutoSparse(dense_ad=AutoForwardDiff())
AutoSparse(dense_ad=AutoForwardDiff(), sparsity_detector=FakeSparsityDetector(), coloring_algorithm=FakeColoringAlgorithm())
```

gdalle: I have added this warning:
```
#=
The following tests are only for visual assessment of the printing behavior.
They do not correspond to proper use of ADTypes constructors.
Please refer to the docstrings for that.
=#
```

wsmoses (approved):

PR #83, opened 2024-09-04
Make sparsity detection and coloring interfaces public

### LINKED ISSUE #82: Export sparsity detection interface
The [interface for sparsity detection](https://sciml.github.io/ADTypes.jl/stable/#Sparsity-detector) should be exported, or at least marked as `public`.
The exported symbols are very unlikely to cause name collisions:
* `jacobian_sparsity`
* `hessian_sparsity`
* `AbstractSparsityDetector`
* `NoSparsityDetector`

The same applies to the matrix coloring interface as well.

adrhill:
Closes #82.


src/ADTypes.jl
```diff
@@ -46,9 +47,21 @@ export AutoChainRules,
        AutoTapir,
        AutoTracker,
        AutoZygote
-
 export AutoSparse
 
+# Sparsity detection
+export AbstractSparsityDetector
```
gdalle, line 53: don't export this, most users won't need it

src/ADTypes.jl
```diff
@@ -46,9 +47,21 @@ export AutoChainRules,
...
+export jacobian_sparsity, hessian_sparsity
+@compat public sparsity_detector
+@compat public NoSparsityDetector
+@compat public KnownJacobianSparsityDetector, KnownHessianSparsityDetector
+
+# Matrix coloring
+export AbstractColoringAlgorithm
```
gdalle, line 60: don't export this, most users won't need it

Project.toml
```diff
@@ -19,6 +18,7 @@ ADTypesEnzymeCoreExt = "EnzymeCore"
 
 [compat]
 ChainRulesCore = "1.0.2"
+Compat = "3.47, 4.10"
```
gdalle, line 21: are we okay with this additional dependency @ChrisRackauckas?

gdalle (changes requested):
thanks!

Project.toml
gdalle: https://discourse.julialang.org/t/what-does-the-new-public-keyword-actually-do/108975/12?u=gdalle

src/ADTypes.jl
```diff
@@ -46,9 +47,21 @@ export AutoChainRules,
        AutoTapir,
        AutoTracker,
        AutoZygote
-
 export AutoSparse
```
gdalle, line 50: `dense_ad` is also public

src/ADTypes.jl
```diff
@@ -6,6 +6,7 @@
 module ADTypes
 
 using Base: @deprecate
+using Compat: @compat # for `public`
```
gdalle, line 9: There are other public names: everything that belongs to the docs page https://sciml.github.io/ADTypes.jl/stable/ is public

gdalle (changes requested):
While we're at it, we may as well do them all

Project.toml
adrhill: I can also copy-and-paste the relevant code from Compat.jl if you want to avoid the 4 stdlib deps Compat.jl has: https://github.com/JuliaLang/Compat.jl/blob/8819abe252579b610cb48c5ad83d51b45a90ddba/Project.toml#L5-L9

src/ADTypes.jl
adrhill: 1. I think the first use-case is quite similar the one of `AbstractADType`: writing an interface that allows users to specify the sparsity detector

2. It's mentioned in the docstrings of the exported interface functions:
  ```markdown
    jacobian_sparsity(f, x, sd::AbstractSparsityDetector)::AbstractMatrix{Bool}
    jacobian_sparsity(f!, y, x, sd::AbstractSparsityDetector)::AbstractMatrix{Bool}
  
  Use detector sd to construct a (typically sparse) matrix S describing the pattern of nonzeroes in the Jacobian of f (resp. f!) applied
  at x (resp. (y, x)).
  ```

3. It makes it easy for users to find sparsity detectors:
  ```Julia-repl
  julia> subtypes(AbstractSparsityDetector)
  3-element Vector{Any}:
   ADTypes.KnownHessianSparsityDetector
   ADTypes.KnownJacobianSparsityDetector
   ADTypes.NoSparsityDetector
  ```
  
4. It makes stack-traces from method errors a bit shorter.


I'm not sure what is being gained by not exporting this.

gdalle: I think exports should be the stuff that people use all the time, and I don't see `AbstractSparsityDetector` being used often. Even for people who wonder which sparsity detectors are available, they'll probably check out the docstring of `jacobian_sparsity` before they remember they can use `subtypes` and other discovery tooling

adrhill: >  Even for people who wonder which sparsity detectors are available, they'll probably check out the docstring of `jacobian_sparsity` before they remember they can use subtypes and other discovery tooling

The docstring of `jacobian_sparsity` is my point 2 above. There are no references of any sparsity detectors in that docstring, only of a (currently unexpected) `AbstractSparsityDetector`. The next logical step for a user is to type `? AbstractSparsityDetector` in the REPL and get an error:

```Julia-repl
julia> using ADTypes

help?> AbstractSparsityDetector
search:

Couldn't find AbstractSparsityDetector
Perhaps you meant AbstractVector
  No documentation found.

  Binding AbstractSparsityDetector does not exist.
```

gdalle: But I don't feel strongly about it so I'll let someone else make the call.

src/ADTypes.jl
adrhill: Added in f8d1397, unless I missed something.

src/ADTypes.jl
gdalle: fair enough, let's keep it

Project.toml
gdalle: Or the shorter version
```julia
VERSION >= v"1.11.0-DEV.469" && eval(Meta.parse("public foo, bar"))
```

Straight from the horse's mouth: https://discourse.julialang.org/t/is-compat-jl-worth-it-for-the-public-keyword/119041/2?u=gdalle

adrhill: Oh god... 😅

adrhill: Removed the dependency on Compat in aae032a.

adrhill: I tried 
```Julia
if VERSION >= v"1.11.0-DEV.469"
    eval(Meta.parse("begin
        # Automatic Differentiation
        public AbstractMode
        public ForwardMode, ReverseMode, ForwardOrReverseMode, SymbolicMode
        public mode

        # Sparse Automatic Differentiation
        public Auto
        public dense_ad

        # Sparsity detection
        public sparsity_detector
        public NoSparsityDetector
        public KnownJacobianSparsityDetector
        public KnownHessianSparsityDetector

        # Matrix coloring
        public coloring_algorithm    
        public NoColoringAlgorithm
    end"))
end
```
but the parser seems to have trouble with this, throwing a `ParseError`:
```
ERROR: ParseError:
# Error @ REPL[27]:20:7
    public NoColoringAlgorithm
#     ┌─────────────────────────
end"))    eval(Meta.parse("begin
    # Automatic Differentiation
⋮
    public NoColoringAlgorithm
    end"))
#────────┘ ── extra tokens after end of expression
Stacktrace:
 [1] top-level scope
   @ none:1
```

adrhill: https://discourse.julialang.org/t/is-compat-jl-worth-it-for-the-public-keyword/119041/5

adrhill: Let's wait a bit until the discussion on discourse has settled.

adrhill: The macro-based approach is much more readable than `eval(Meta.parse(...))`.

gdalle:
Since the discussion in https://discourse.julialang.org/t/is-compat-jl-worth-it-for-the-public-keyword/119041/ has come to a close, I think I'd rather put all the `public` declarations for ADTypes in one file, which makes the following approach the most appealing:
```julia
if VERSION > ...
    include("public.jl")
end
```

adrhill:
Looking at the "likes" on Discourse, it looks to me like most people favor Mason's macro-based solution, which is what is implemented here as of 1cd4162. The disadvantages of a separate `src/public.jl` file have been discussed on Discourse as well.

Have you taken a look at the current diff? I think it's simple, succinct and readable.

adrhill:
I'm obviously biased towards my own code, but I like the clean overview it provides in the main file, separated by use-case:

```Julia

# Automatic Differentiation
export AbstractADType
export AutoChainRules,
       AutoDiffractor,
       AutoEnzyme,
       AutoFastDifferentiation,
       AutoFiniteDiff,
       AutoFiniteDifferences,
       AutoForwardDiff,
       AutoModelingToolkit,
       AutoPolyesterForwardDiff,
       AutoReverseDiff,
       AutoSymbolics,
       AutoTapir,
       AutoTracker,
       AutoZygote
@public AbstractMode
@public ForwardMode, ReverseMode, ForwardOrReverseMode, SymbolicMode
@public mode
@public Auto

# Sparse Automatic Differentiation
export AutoSparse
@public dense_ad

# Sparsity detection
export AbstractSparsityDetector
export jacobian_sparsity, hessian_sparsity
@public sparsity_detector
@public NoSparsityDetector
@public KnownJacobianSparsityDetector
@public KnownHessianSparsityDetector

# Matrix coloring
export AbstractColoringAlgorithm
export column_coloring, row_coloring, symmetric_coloring
@public coloring_algorithm
@public NoColoringAlgorithm
```

gdalle:
In the spirit of maintainability, I would like to avoid macros whenever I can. But I don't feel very strongly about this so if others prefer this way I'll yield.

adrhill:
As you know, I generally agree with that sentiment, but in this case the macro simply returns `Expr(:public, args...)`.

gdalle (changes requested):
I added tests on 1.11 in #84. Can you merge `main` into this PR and maybe test that everything works on 1.11, i.e. that the symbols are indeed marked public?
Also for some reason I didn't see code coverage?

adrhill:
> Also for some reason I didn't see code coverage?

Looks like it wasn't triggered on previous PRs either. The [codecov PR page](https://app.codecov.io/gh/SciML/ADTypes.jl/pulls) looks odd as well, with all PRs being untitled. Maybe just the usual codecov bugginess?

gdalle:
Fair enough. Are the symbols correctly marked on 1.11?

adrhill:
How do I check this? By looking at the docstrings in REPL help mode?

gdalle:
I think the easiest is to add a test that only runs on 1.11 and that verifies whether the unexported Symbols we marked `public` are part of `names(ADTypes)`.
https://discourse.julialang.org/t/what-does-the-new-public-keyword-actually-do/108975/3?u=gdalle

gdalle (approved):
perfectly balanced, like all things should be

PR #89, opened 2024-09-25
Add AutoMooncake type

willtebbutt:
## Checklist

## Additional context

`AutoTapir` needs to be deprecated in favour of `AutoMooncake`, because Tapir.jl has been renamed to Mooncake.jl.

I could do with some advice on the appropriate way to deprecate `AutoTapir` -- should I just add the constructor to the `legacy.jl` file, or is there something else that I should do?


gdalle:
Can we have a default setting so that people don't need to dig into the `Mooncake.Config` docs? Like `AutoMooncake(; config=nothing)`?

willtebbutt:
We could do that. I'd prefer to prevent people from writing `AutoMooncake()` -- they should at least have to write `AutoMooncake(; config=nothing)` -- in order to ensure that people are aware that they have some configuration options.

gdalle:
Sounds good, let's do that. As a user of ADTypes it's always a bit painful to go grab a specific type from a package to configure backends (like `AutoZygote(; config=Zygote.ZygoteRuleConfig())`)

gdalle:
Ready for review?

docs/src/index.md
```diff
@@ -39,9 +39,10 @@ AutoGTPSA
 ### Reverse mode
 
 ```@docs
+AutoMooncake
 AutoReverseDiff
-AutoTapir
 AutoTracker
+AutoTapir
```
gdalle, line 45: We could put this in a level-2 section
```
## Deprecated
```
at the bottom of the page

src/dense.jl
```diff
@@ -227,6 +227,30 @@ function Base.show(io::IO, backend::AutoGTPSA{D}) where {D}
     print(io, ")")
 end
 
+"""
+    AutoMooncake
+
+Struct used to select the [Mooncake.jl](https://github.com/compintell/Mooncake.jl) backend
+for automatic differentiation.
```
gdalle, line 234:
```suggestion
Struct used to select the [Mooncake.jl](https://github.com/compintell/Mooncake.jl) backend for automatic differentiation.
```

src/dense.jl
```diff
@@ -227,6 +227,30 @@ function Base.show(io::IO, backend::AutoGTPSA{D}) where {D}
...
+
+# Constructors
+
+    AutoMooncake(; config)
+
+# Fields
+
+  - `config`: either `nothing` or an instance of `Mooncake.Config` -- see the docstring for
+    `Mooncake.Config` for more information. `AutoMooncake(; config=nothing)` is equivalent
+    to `AutoMooncake(; config=Mooncake.Config())`, i.e. the default configuration.
```
gdalle, line 246:
```suggestion
  - `config`: either `nothing` or an instance of `Mooncake.Config` -- see the docstring of `Mooncake.Config` for more information. `AutoMooncake(; config=nothing)` is equivalent to `AutoMooncake(; config=Mooncake.Config())`, i.e. the default configuration.
```

src/dense.jl
```diff
@@ -323,26 +347,10 @@ mode(::AutoSymbolics) = SymbolicMode()
...
-
-Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
-
-# Constructors
-
-    AutoTapir(; safe_mode=true)
-
-# Fields
-
-  - `safe_mode::Bool`: whether to run additional checks to catch errors early. While this is
```
gdalle, line 336: Maybe we could preserve this part of the old docstring, so that future maintainers of ADTypes know it should just be left that way (rather than deleted)?

src/dense.jl
```diff
@@ -323,26 +347,10 @@ mode(::AutoSymbolics) = SymbolicMode()
...
-    on by default to ensure that users are aware of this option, you should generally turn
-    it off for actual use, as it has substantial performance implications.
-    If you encounter a problem with using Tapir (it fails to differentiate a function, or
-    something truly nasty like a segfault occurs), then you should try switching `safe_mode`
-    on and look at what happens. Often errors are caught earlier and the error messages are
-    more useful.
+This ADType is deprecated. `AutoMooncake` should be used instead.
```
gdalle, line 350:
```suggestion
!!! danger

    `AutoTapir` is deprecated following a package renaming, please use [`AutoMooncake`](@ref) instead.
```

src/legacy.jl
```diff
@@ -36,3 +36,10 @@ function AutoModelingToolkit(; obj_sparse::Bool = false, cons_sparse::Bool = fal
         :AutoModelingToolkit; force = false)
     return mtk_to_symbolics(obj_sparse, cons_sparse)
 end
+
+function AutoTapir(; safe_mode=true)
+    Base.depwarn(
+        "AutoTapir is deprecated in favour of AutoMooncake.", :AutoTapir; force=false
```
gdalle, line 42:
```suggestion
        "`AutoTapir` is deprecated in favour of `AutoMooncake`.", :AutoTapir; force=false
```

src/dense.jl
gdalle: But we put the big red warning right on top

test/legacy.jl
```diff
@@ -68,3 +68,7 @@ end
     ad = @test_deprecated AutoReverseDiff(true)
     @test ad.compile
 end
+
+@testset "AutoTapir" begin
+    @test_deprecated AutoTapir()
```
gdalle, line 73:
```suggestion
    @test_deprecated AutoTapir()
    @test_deprecated AutoTapir(; safe_mode=false)
```

gdalle (changes requested):
Thanks for this contribution! Most of my remarks are esthetics, fundamentally it is good to go

gdalle (approved):

PR #94, opened 2024-11-13
feat: add ConstructionBaseExt to allow Setfield and Functors support

avik-pal:
## Checklist

- [ ] ~All documentation related to code changes were updated~
- [ ] ~Any new documentation only uses public API~
  
## Additional context

Functors v0.5 started defaulting to recurse into all types, which means we need to be able to reconstruct types based on ConstructionBase. See https://github.com/LuxDL/Lux.jl/actions/runs/11825400363/job/32949087528?pr=1069#step:6:50


ChrisRackauckas (approved):

gdalle:
Please don't merge before I have a chance to review

ChrisRackauckas:
Please review now then? It's very uncontroversial.

ext/ADTypesConstructionBaseExt.jl
```diff
@@ -0,0 +1,18 @@
+module ADTypesConstructionBaseExt
+
+using ADTypes: AutoEnzyme, AutoForwardDiff, AutoPolyesterForwardDiff
```
gdalle, line 3: Why do you only consider these three types?

test/misc.jl
```diff
@@ -61,3 +61,36 @@ for backend in [
 ]
     println(backend)
 end
+
+using Setfield
+
+@testset "Setfield compatibility" begin
```
gdalle, line 67: I haven't encountered Setfield and ConstructionBase before. IIUC these tests would have failed without the extension?

gdalle (changes requested):

ext/ADTypesConstructionBaseExt.jl
ChrisRackauckas: Why would the other ones need to be specialized?

test/misc.jl
ChrisRackauckas: Some would have passed of course, but there are a few with required type parameters which would have failed

ext/ADTypesConstructionBaseExt.jl
gdalle: I don't know, as stated below I am not familiar with the packages being discussed here. I just wondered if focusing only on Enzyme and ForwardDiff was a shortcut to answer only @avik-pal's needs or if they are truly the only backends for which such an overload is necessary.

ChrisRackauckas: It always works with the default constructors, and those are the ones for which it needs to be specialized.

[It seems really weird to say that no one can merge until you review but then say you don't know the packages involved in the extension.....?]

avik-pal: https://juliaobjects.github.io/ConstructionBase.jl/dev/#type-tips. others will work without a custom dispatch

test/misc.jl
avik-pal: yes

ext/ADTypesConstructionBaseExt.jl
gdalle: > [It seems really weird to say that no one can merge until you review but then say you don't know the packages involved in the extension.....?]

On the contrary, it seems rather logical? Due to DI, I am pretty much the only person in the ecosystem using the integrality of ADTypes, so I would appreciate having time to review meaningful changes to this package. I mentioned it because I know you like to merge PRs rather quickly once you approve them (often on a scale of minutes).

The fact that I don't know the packages involved is not a counter argument to my reviewing. On the contrary, I think it's fair that I take some time to understand what happens here. If I added an extension to some SciML package depending on something you'd never heard of, you'd probably want to understand too.

I'll take a look at the link @avik-pal gave, and give a review tomorrow.

gdalle:
> Please review now then? It's very uncontroversial.

It is currently 23:18 where I live. I don't think this change is so urgent that it cannot wait 24h.

ext/ADTypesConstructionBaseExt.jl
avik-pal: To Chris's point, this is not going to break anyone's code and is a simple enough change that shouldn't need that long to get merged.

It is blocking a major update to Lux which in turn blocks a correctness update to Flux.

gdalle: To put things in perspective: 24h is a ridiculously short time for something to get merged. Just because Chris often reviews and merges PRs within the hour doesn't mean everyone is held to the same standards.

You'll have a review by me tomorrow, where I make sure that nothing in DI is affected. I endeavour to do the same for all ADTypes PRs. If you don't think my input is necessary, feel free to merge. But I do believe that the special integration of ADTypes and DI would justify that, as a general rule, I am given more than a few minutes to review PRs to this repo. This is especially relevant considering time zone differences (I'm in Europe) and the fact that, unlike Chris, I'm not working full time on SciML.

ChrisRackauckas: > But I do believe that the special integration of ADTypes and DI would justify that, as a general rule, I am given more than a few minutes to review PRs to this repo. 

I moved the full answer to https://github.com/SciML/ADTypes.jl/issues/95, but in summary, let's make sure you get more tests so you're comfortable here. That's a system problem not a people problem.

But also, if you don't use SetField, why would that break DI? I would presume that if you don't use SetField, then DI doesn't use SetField, so adding SetField support shouldn't even effect DI? I don't see the logic thread here beyond paranoia of missing tests.

>  This is especially relevant considering time zone differences (I'm in Europe) and the fact that, unlike Chris, I'm not working full time on SciML.

? Interesting comment 😅 . For the record, I'm also in Europe (ish) and I have (a few) full time jobs on proprietary codebases that come before any academic work (with open source work part of the academic work). Though I wouldn't recommend

gdalle: > I moved the full answer to https://github.com/SciML/ADTypes.jl/issues/95, but in summary, let's make sure you get more tests so you're comfortable here. That's a system problem not a people problem.

That's a fair answer, I'll react over there. Thank you.

> But also, if you don't use SetField, why would that break DI? I would presume that if you don't use SetField, then DI doesn't use SetField, so adding SetField support shouldn't even effect DI? I don't see the logic thread here beyond paranoia of missing tests.

If it were just an extension indeed there would be zero probability of breaking DI. But as soon as constructors are modified, which is the case here, I think that DI and other packages are more likely to break. The main reason for that is #96.

> For the record, I'm also in Europe (ish) and I have (a few) full time jobs on proprietary codebases that come before any academic work (with open source work part of the academic work).

Point taken, sorry for assuming.

> Though I wouldn't recommend my hours, but that's a topic for another day.

But that's exactly the point, isnt't it? Even from you, no one should expect a review within the hour. You're also entitled to sleep and food and maybe even hobbies. Regardless of the root cause (which we discuss in #95), if 24h doesn't seem like a reasonable reviewing delay, then we have a problem as a community because that really doesn't scale.

src/dense.jl
```diff
@@ -67,9 +67,11 @@ struct AutoEnzyme{M, A} <: AbstractADType
     mode::M
 end
 
+AutoEnzyme{A}(mode::M) where {M, A} = AutoEnzyme{M, A}(mode)
```
gdalle, line 70: Apart from the ConstructionBase stuff (on which I might have to trust you, and that's okay), my main concern with this PR is here. Usually, constructors with only a subset of type parameters consider type parameters in the right order. This modification (unlike the one on `AutoForwardDiff`) is in conflict with the general convention.

Given a type `AutoEnzyme{T1,T2}`, if there is a shortcut `AutoEnzyme{T}`, I would expect `T` to refer to `T1` and not `T2`. This would also be more coherent with the way that [`UnionAll` types work](https://docs.julialang.org/en/v1/devdocs/types/#UnionAll-types). Otherwise, the two following uses are not in agreement:
```julia
backend = AutoEnzyme{T}(stuff) # T is T2
backend isa AutoEnzyme{T} # T is T1
```
What do you think?

gdalle (changes requested):

ext/ADTypesConstructionBaseExt.jl
gdalle: See #95 for the discussion, no need to continue in two places. TLDR: Chris is right and I should chill a little ^^

src/dense.jl
gdalle: That hack is also what I had in mind, and I think it is better if we consider #96, i.e. the fact that user-unfriendly constructors `AutoSomething{T1,T2}(arg1, arg2)` may become part of the public API.

src/dense.jl
```diff
@@ -69,7 +69,7 @@ end
 
 function AutoEnzyme(;
         mode::M = nothing, function_annotation::Type{A} = Nothing) where {M, A}
-    return AutoEnzyme{M, A}(mode)
+    return AutoEnzyme{A}(mode)
```
gdalle, line 72:
```suggestion
    return AutoEnzyme{M, A}(mode)
```

gdalle:
You can go ahead and merge, I trust Chris' reviey

PR #109, opened 2025-05-31
Add DI warning to AutoEnzyme

gdalle:
## Checklist

## Additional context

Calling Enzyme through DI can introduce errors or slowdowns compared to Enzyme's native API. However, end users are unlikely to check out either the DI docs or the Enzyme docs when investigating what happened, especially when derivatives are deep in the call stack.
On the other hand, end users usually provide the `AutoEnzyme` object to whatever package they are using. Thus, adding a warning there seems like a good way to raise awareness.

@wsmoses what do you think? You could also write a page for the Enzyme docs with examples of what can go wrong when going through DI, and we could link it from here and from the DI docs.

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -45,6 +45,11 @@ Struct used to select the [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl) bac
 
 Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
 
+!!! warning
+    `AutoEnzyme` can be used by [DifferentiationInterface.jl](https://github.com/JuliaDiff/DifferentiationInterface.jl) to access a restricted subset of Enzyme.jl functionality.
+    If your code or one of its dependencies computes derivatives through DifferentiationInterface.jl, you might experience slowdowns or errors that would not happen with Enzyme.jl's native API.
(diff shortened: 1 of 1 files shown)


src/dense.jl
```diff
@@ -45,6 +45,11 @@ Struct used to select the [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl) bac
 
 Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
 
+!!! warning
+    `AutoEnzyme` can be used by [DifferentiationInterface.jl](https://github.com/JuliaDiff/DifferentiationInterface.jl) to access a restricted subset of Enzyme.jl functionality.
```
wsmoses, line 49: This feels like the wrong place to have it, since sciml uses ADTypes (created the package), and calls Enzyme directly

wsmoses: Yeah but this still feels like the wrong place for this to happen, since ADTypes often directly calls enzyme and doesn’t go through DI.

As mentioned in the other thread I’m working on trying to make the errors throw only if called within DI

gdalle: I thought this would be a good addition for the many packages who do call Enzyme through DI, without hurting the others, but I'll leave it up to you. Feel free to revive this PR if you change your mind, I'm setting it as a draft for the time being.

PR #110, opened 2025-06-05
feat: Add forward mode Mooncake

gdalle:
## Checklist

## Additional context

This PR adds the `AutoMooncakeForward` type, in anticipation of https://github.com/chalk-lab/Mooncake.jl/pull/389 landing. It is different from the existing `AutoMooncake` because otherwise we would have had to replace `mode(::AutoMooncake) = ForwardMode()` with `mode(::AutoMooncake) = ForwardOrReverseMode()`, which can be considered breaking.

cc @willtebbutt


src/ADTypes.jl
```diff
@@ -43,6 +43,7 @@ export AutoChainRules,
        AutoGTPSA,
        AutoModelingToolkit,
        AutoMooncake,
+       AutoMooncakeForward,
```
yebai, line 46: For completeness, I suggest we consider introducing `AutoMooncakeBackward`, then alias `AutoMooncake = AutoMooncakeBackward`.

gdalle: I think having two names for the same thing introduces unneeded confusion. We can remove `AutoMooncakeForward` at the next breaking release of ADTypes if there ever is one

ChrisRackauckas:
Though I'd be willing to do that breaking change "soon" so that it's before Mooncake adoption is big.

PR #115, opened 2025-07-27
Remove Julia v1.9 get_extension compatibility code

ChrisRackauckas:
## Description

This PR removes the compatibility checks for `isdefined(Base, :get_extension)` since all SciML packages now require Julia v1.10+, where package extensions are built-in.

## Changes

- Removed unnecessary version checks in extension loading code
- Simplified extension imports by removing conditional logic
- Cleaned up obsolete compatibility code

## Context

As identified in the SciML-wide analysis, all SciML packages have moved to requiring Julia v1.10+ as the minimum version. This makes the compatibility code for checking whether package extensions are available redundant.

The `isdefined(Base, :get_extension)` checks were used to support Julia v1.9 where extensions were loaded differently. Since we no longer support v1.9, this code can be safely removed.

## Testing

- [ ] Package tests pass locally
- [ ] No changes to functionality, only removal of version checks

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,36 +1,46 @@
-name: Downgrade
+name: CI
 on:
-  pull_request:
-    branches:
-      - master
-    paths-ignore:
-      - 'docs/**'
   push:
     branches:
-      - master
-    paths-ignore:
-      - 'docs/**'
+      - main
+    tags: ['*']
+  pull_request:
+concurrency:
+  # Skip intermediate builds: always.
+  # Cancel intermediate builds: only if it is a pull request build.
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


ChrisRackauckas:
I'll get a script going to find any other stragglers. 

This isn't triggering package tests though?

ChrisRackauckas:
This was a good test of the new downgrade 😅 sorry that rollout was a bit of a mess.

@thazhemadam can you help me get downgrade into the https://github.com/SciML/.github/tree/master/.github/workflows and roll that out globally? I think we should just make the normal CI have an option to error on depwarn and turn it off by default just to get the rest of the rollout around, as I'm surprised this repo isn't on the global CI yet. Also, I'm surprised not to see downstream on here, but now we should just wait for the rollout

PR #117, opened 2025-07-28
Add :Mooncake to Auto

MilesCranmer:


MilesCranmer:
Thanks. Would it be ok to make a new release too?

Issue #123, opened 2025-08-15
Export AutoEnzymeForward / AutoEnzymeReverse


wsmoses:
Makes it easier for folks to use things, and mirrors other tools like Mooncake

gdalle:
Actually we were thinking of doing the opposite and removing `AutoMooncakeForward` in the next release, in favor of a `mode` kwarg like in the `AutoEnzyme` API. The only reason why it was introduced that way was to avoid a breaking change. What's wrong with the current `AutoEnzyme(; mode=Enzyme.Forward)`?

Related:

- https://github.com/SciML/ADTypes.jl/pull/110
- https://github.com/SciML/ADTypes.jl/issues/113

gdalle:
See discussion in #114

Issue #113, opened 2025-07-22
Make it unnecessary to load Enzyme and FiniteDifferences to specify AutoEnzyme and AutoFiniteDifferences


PTWaade:
Dear ADTypes maintainers,

Thank you for your work!

This issue is just to request that the `AutoEnzyme` and `AutoFIniteDifferences` constructors be updated so that it is not necessary to have Enzyme and FiniteDifferences loaded in order to construct them.
The reason this is currently necessary, is because AutoEnzyme requires the `Forward` and `Reverse` types to specify whether Enzyme is used in forward or reverse mode, and also requires `set_runtime` if it is to be used with Turing (which is my use-case); similarly, AutoFiniteDifferences requires importing `central_fdm` from FiniteDifferences to be constructed.

There are two reasons that it would be useful to not have to load the packages themselves to use the AutoAD constructors. Firstly, it strikes me as much easier to read and use, especially by people who just want to "turn on" their choice of AD in e.g. a Turing model, if things like forward and reverse mode can just be specified with a flag or an otherwise normal keyword argument. Secondly, and more importantly, in my case I would like to use AutoEnzyme as a default argument for a function, but I cannot do that without making Enzyme a hard dependency of my package.
Since the AutoEnzyme functionality is optional, I would like to avoid making it a hard dependency, and that users should just import Enzyme to use it.
I think in Turing's case there is the same problem, but with wanting to use FiniteDifferences as default, @penelopeysm tells me.

I'll note that this works fine for other constructors, such as AutoForwardDiff (whcih is currently used as default from Turing because of this), AutoReverseDiff and AutoMooncake. 
I also discussed this with Enzyme, but they said that this isn't an issue they can fix, since it is in your package.

There may also be other AutoAD constructors in your (very useful!) list that have the same limitation; I haven't checked them all, Enzyme and FiniteDifferences are just the two I ran into. My suggestion would be to make it a general policy, however, that the ADTypes constructors do not require loading the corresponding AD package in order to be constructed.

There may of course be many good reasons that I am ignorant about for why using AD package specific types and functions in your constructors is useful, or that changing it now is complicated - so forgive me if this is a stupid question. 

And thank you - for the consideration and time, and again, for your work :)

PTWaade:
Hey @gdalle 

Thank you for your thoughtful response, it's appreciated! 

[...]

PR #124, opened 2025-08-15
feat: add chunksize parameter to `AutoEnzyme`

### LINKED ISSUE #85: Adapt to Enzyme 0.13
Enzyme has a pending breaking release upcoming.

One notable addition is runtime activity being part of the mode now, we well as (potentially) moving whether the primal is returned as part of the mode.

Runtime activity probably makes sense to add to ADTypes, whether to return the primal, also potentially (though the latter hasn't even landed yet)

### LINKED ISSUE #114: More parameters for `AutoEnzyme`
I think it would make sense to add more parameters to `AutoEnzyme`, like:

- `deferred = true / false` to use `autodiff_deferred` instead of `autodiff`
- `chunksize` to imitate the `AutoForwardDiff` API and allow selecting a chunk size for Jacobians and Hessians

@wsmoses @vchuravy what do you think?

Related:

- https://github.com/JuliaDiff/DifferentiationInterface.jl/issues/819#issuecomment-3070516150

gdalle:
## Checklist

## Additional context

Fixes #85, fixes #114

Note that these changes rely on downstream packages (mostly DI) to interpret them correctly when reconstructing an Enzyme mode object before differentiation. Between the release of ADTypes v1.18 and the release of the corresponding DI patch, the new backend parameters introduced here will have zero effect. That's not a great situation but ADTypes has no way to require a future DI version before introducing these changes. The real solution to prevent such gaps would probably be to merge ADTypes into DI, so that the semantics are defined at the same time as the backends.

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -39,7 +39,7 @@ struct AutoDiffractor <: AbstractADType end
 mode(::AutoDiffractor) = ForwardOrReverseMode()
 
 """
-    AutoEnzyme{M,A,R,C}
+    AutoEnzyme{M,A,C}
 
 Struct used to select the [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl) backend for automatic differentiation.
 
@@ -50,7 +50,6 @@ Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
     AutoEnzyme(;
         mode::Union{EnzymeCore.Mode,Nothing}=nothing,
         function_annotation::Type{<:Union{EnzymeCore.Annotation,Nothing}}=Nothing,
-        runtime_activity::Union{Bool,Nothing}=nothing,
         chunksize::Union{Int,Float64,Nothing}=nothing,
     )
 
@@ -64,48 +63,40 @@ Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
       + a subtype of `EnzymeCore.Annotation` (like `EnzymeCore.Const` or `EnzymeCore.Duplicated`) to enforce a given annotation
 
       + `Nothing` (the type, not the object) to simply pass `f` and let Enzyme choose the most appropriate annotation
-  - `R=runtime_activity` determines whether [runtime activity](https://enzymead.github.io/Enzyme.jl/stable/faq/#faq-runtime-activity) is activated by Enzyme. It can be:
-
-      + a `Bool` to force a given setting
-
-      + `nothing` to fall back on the setting of the `mode` object
   - `C=chunksize` determines the number of derivatives evaluated simultaneously when computing operators like a Jacobian or a forward-mode gradient. It can be:
 
       + a positive `Int` to fix a constant chunk size
       + `Inf` to pick the maximum chunk size, corresponding to the array length
       + `nothing` to choose a good chunk size automatically
 """
-struct AutoEnzyme{M, A, R, C} <: AbstractADType
+struct AutoEnzyme{M, A, C} <: AbstractADType
     mode::M
 
-    function AutoEnzyme{M, A, R, C}(mode::M) where {M, A, R, C}
-        @assert R isa Union{Nothing, Bool}
+    function AutoEnzyme{M, A, C}(mode::M) where {M, A, C}
         @assert C isa Union{Nothing, Int, Float64}
         if C isa Int
             @assert C > 0
         elseif C isa Float64
             @assert C == Inf
         end
-        return new{M, A, R, C}(mode)
+        return new{M, A, C}(mode)
     end
 end
 
 function AutoEnzyme(;
         mode::M = nothing,
         function_annotation::Type{A} = Nothing,
-        runtime_activity::Union{Nothing, Bool} = nothing,
         chunksize::Union{Nothing, Int, Float64} = nothing
 ) where {M, A}
-    return AutoEnzyme{M, A, runtime_activity, chunksize}(mode)
+    return AutoEnzyme{M, A, chunksize}(mode)
 end
 
 mode(::AutoEnzyme) = ForwardOrReverseMode()  # specialized in the extension
 
-function Base.show(io::IO, backend::AutoEnzyme{M, A, R, C}) where {M, A, R, C}
+function Base.show(io::IO, backend::AutoEnzyme{M, A, C}) where {M, A, C}
     print(io, AutoEnzyme, "(")
     !isnothing(backend.mode) && print(io, "mode=", repr(backend.mode; context = io), ", ")
     !(A <: Nothing) && print(io, "function_annotation=", repr(A; context = io), ", ")
-    !(R === nothing) && print(io, "runtime_activity=", repr(R; context = io), ", ")
     !(C === nothing) && print(io, "chunksize=", repr(C; context = io))
     print(io, ")")
 end
--- a/ext/ADTypesConstructionBaseExt.jl
+++ b/ext/ADTypesConstructionBaseExt.jl
@@ -3,14 +3,14 @@ module ADTypesConstructionBaseExt
 using ADTypes: AutoEnzyme, AutoForwardDiff, AutoPolyesterForwardDiff
 using ConstructionBase: ConstructionBase
 
-struct InternalAutoEnzymeReconstructor{A, R, C} end
+struct InternalAutoEnzymeReconstructor{A, C} end
 
-function InternalAutoEnzymeReconstructor{A, R, C}(mode::M) where {M, A, R, C}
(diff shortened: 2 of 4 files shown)


wsmoses (commented):
What if specifying the runtime activity flag with the auto forward mode created an EnzymeCore.set_runtime_activity(enzymecore.forward), that would mean that downstream users don't change

gdalle:
Downstream code needs to change anyway, if only because of the chunk size.
Indeed I thought of adding a converter inside the EnzymeCore extension, but that would make it necessary for EnzymeCore to be loaded for the `AutoEnzyme(; mode=ADTypes.ForwardMode())` constructor to run. I guess it depends how extreme you want to be about #113 / #123: does your ADTypes constructor need to work before Enzyme is loaded or not?

wsmoses:
Yeah but making sure that the downstream users don't need to reimplement the "to mode" logic is nice -- and especially making sure that it's implemented correctly here.

Is there a reason for not doing that offhand?

gdalle:
The main reason is people wanting to be able to specify the mode without having Enzyme loaded (see #113 or #123). If you really want to do that, then you need to defer the `to_mode` call to the differentiation step.
But in all honesty, that sounds like a bad idea to me anyway. If you're going to use Enzyme, you might as well load it, or rather load EnzymeCore which is very lightweight. Furthermore, having more than one syntax to specify the mode is unneeded complexity. I'd much prefer removing everything from this PR except the chunk size addition, which is not present in the Enzyme mode object and which is consistent with other ADTypes backends.

gdalle:
The middle ground you're suggesting is for people who are ok with loading EnzymeCore before constructing the backend object, but are not okay using the EnzymeCore mode objects, so they want something like this

```julia
using EnzymeCore, ADTypes
backend = AutoEnzyme(mode=ADTypes.ForwardMode())
```

rather than something like that

```julia
using EnzymeCore, ADTypes
backend = AutoEnzyme(mode=EnzymeCore.Forward)
```

which sounds very strange to me

gdalle:
To clarify, I completely agree with you that getting the EnzymeCore mode object as soon as possible and as uniquely as possible is a good thing for correctness. I'm just trying to be the devil's advocate and argue in favor of a request like #113, which I tried to entertain in this PR. But my opinion is that the added complexity is not worth it, and I don't really see a scenario where loading EnzymeCore at backend definition time is a bad thing.

wsmoses:
I think it's mostly making as easy as possible for the user (just seeing the ADTypes package from their end, and not knowing they want to import EnzymeCore in place of enzyme, which seems to be the case in the linked issue).

Also one we can potentially avoid the import two packages problem to register to mode is just depending on EnzymeCore (it is dependency free and rarely changed so shouldn't cause any problems)

gdalle:
To sum up, we have four options:

1. Specify mode with EnzymeCore only (current behavior)
2. Specify mode with EnzymeCore or ADTypes, conversion to EnzymeCore downstream (current state of this PR)
3. Specify mode with EnzymeCore or ADTypes, conversion to EnzymeCore in ADTypes extension
4. Specify mode with EnzymeCore or ADTypes, conversion to EnzymeCore in ADTypes itself with new dependency on EnzymeCore

I agree that 2 is not great for correctness. I'm putting a veto on 4, cause I don't want ADTypes or DI getting a hard dependency on any AD package, no matter how lightweight. This leaves 1 or 3. My preference would be to keep 1, precisely because I deem it more simple to have just one syntax, but if you think your users would prefer 3 with the additional ADTypes mode and the conversion in EnzymeCoreExt, I'm happy to adapt the PR. Your call.

gdalle:
@wsmoses I just remembered a reason why options 3 and 4 are not possible anyway: when the user selects `AutoEnzyme(; mode = nothing)`, this means that we need to pick the best mode _depending on the differentiation operator_ that gets used. For `DI.pushforward`, it will be a forward mode, but for `DI.gradient`, it will be a reverse mode (and I assume other packages like Optimization.jl do something similar). This kind of choice needs to happen downstream, so the conversion to an EnzymeCore object needs to happen downstream as well (because we cannot convert `nothing` to an `EnzymeCore.mode` without knowing the operator to apply).

gdalle:
In other words, I think this PR is the right version, and I'd appreciate other reviews

src/dense.jl
```diff
@@ -39,46 +39,77 @@ struct AutoDiffractor <: AbstractADType end
...
+      + a `Bool` to force a given setting
 
-      + an object subtyping `EnzymeCore.Mode` (like `EnzymeCore.Forward` or `EnzymeCore.Reverse`) if a specific mode is required
-      + `nothing` to choose the best mode automatically
+      + `nothing` to fall back on the setting of the `mode` object
+  - `C=chunksize` determines the number of derivatives evaluated simultaneously when computing operators like a Jacobian or a forward-mode gradient. It can be:
+
+      + a positive `Int` to fix a constant chunk size
+      + `Inf` to pick the maximum chunk size, corresponding to the array length
+      + `nothing` to choose a good chunk size automatically®
```
gdalle, line 77:
```suggestion
      + `nothing` to choose a good chunk size automatically
```

wsmoses:
I actually dislike that convention, because it means that people don't know what to expect from downstream packages (e.g. maybe someone chooses forward vs reverse, and something fails unexpectedly). Similarly here you end up in the situation where you have multiple conflicting ways of specifying things -- leading to further confusion/ambiguity.

Since I think I would push for 4 or 3, and veto 2; and you would push for 2 and veto 4, I think that means we stick with 1.

gdalle:
> I actually dislike that convention, because it means that people don't know what to expect from downstream packages (e.g. maybe someone chooses forward vs reverse, and something fails unexpectedly). Similarly here you end up in the situation where you have multiple conflicting ways of specifying things -- leading to further confusion/ambiguity.

Yeah, the default `AutoEnzyme()` is a compromise. It's open to downstream interpretation, but at least it doesn't expect beginners to know what forward and reverse modes are, or which one is (usually) best for their given application. In any case, it's probably too breaking to remove.

> Since I think I would push for 4 or 3, and veto 2; and you would push for 2 and veto 4, I think that means we stick with 1.

Alright then, I removed the ADTypes mode specification, but I left runtime activity and chunksize choices. In particular, this will allow people to use `AutoEnzyme(; runtime_activity=true)` in e.g. Turing and have DI automatically pick `set_runtime_activity(Reverse)` as the mode when computing gradients, which is a nice QoL improvement.

If you think this is good to go, we can merge and I can do the DI follow up.

gdalle:
Fair enough, I removed it in the last commit. Now this only adds a chunk size and changes nothing else

gdalle:
Gentle bump @wsmoses

src/dense.jl
```diff
@@ -39,46 +39,65 @@ struct AutoDiffractor <: AbstractADType end
...
-struct AutoEnzyme{M, A} <: AbstractADType
+struct AutoEnzyme{M, A, C} <: AbstractADType
     mode::M
+
+    function AutoEnzyme{M, A, C}(mode::M) where {M, A, C}
+        @assert C isa Union{Nothing, Int, Float64}
+        if C isa Int
+            @assert C > 0
+        elseif C isa Float64
+            @assert C == Inf
```
wsmoses, line 80: we should give a better error message here

src/dense.jl
```diff
@@ -39,46 +39,65 @@ struct AutoDiffractor <: AbstractADType end
...
-# Fields
+      + a subtype of `EnzymeCore.Annotation` (like `EnzymeCore.Const` or `EnzymeCore.Duplicated`) to enforce a given annotation
 
-  - `mode::M` determines the autodiff mode (forward or reverse). It can be:
+      + `Nothing` (the type, not the object) to simply pass `f` and let Enzyme choose the most appropriate annotation
+  - `C=chunksize` determines the number of derivatives evaluated simultaneously when computing operators like a Jacobian or a forward-mode gradient. It can be:
 
-      + an object subtyping `EnzymeCore.Mode` (like `EnzymeCore.Forward` or `EnzymeCore.Reverse`) if a specific mode is required
-      + `nothing` to choose the best mode automatically
+      + a positive `Int` to fix a constant chunk size
```
wsmoses, line 68: I kind of wonder if a chunk size of 0 here would be a good way to represent maximum chunk size

wsmoses: reading that, I still don't understand why a zero chunksize is semantically meaningful?

gdalle: Hi Billy, just following up on this, do you want to add that setting to EnzymeCore?

gdalle: See https://github.com/EnzymeAD/Enzyme.jl/pull/2659

PR #127, opened 2025-11-07
Add AutoReactant{AutoEnzyme}

wsmoses:
It's not really an AD backend, but this can help make things easier for folks to use throughout the ecosystem.

I'm not quite sure this is the right place/abstraction though, alternatively we can add bool flags to turn on reactant/etc as applicable throughout

cc @avik-pal

### DIFF
--- a/src/dense.jl
+++ b/src/dense.jl
@@ -98,7 +98,7 @@ Defined by [ADTypes.jl](https://github.com/SciML/ADTypes.jl).
 
   - `mode::M` specifies the parameterization of differentiation.  It can be:
 
-      + an AutoEnzyme object if a specific mode is required
+      + an [`AutoEnzyme`](@ref) object if a specific mode is required
       + `nothing` to choose the best mode automatically
 """
 struct AutoReactant{M<:AutoEnzyme} <: AbstractADType


gdalle:
Can you give an example of where this would simplify things? Lux's API?

gdalle:
In that case it might be time to revive my initial attempt https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/325 for the general case.

I'm not fundamentally opposed to the principle of the current PR (haven't reviewed yed) but there is a documentation challenge to convey what Reactant is, how it is not an AD backend but it can instead combine with AD backends

wsmoses:
> In that case it might be time to revive my initial attempt [JuliaDiff/DifferentiationInterface.jl#325](https://github.com/JuliaDiff/DifferentiationInterface.jl/pull/325) for the general case.

Yeah go for, I think that is just waiting for someone to bring it over the line

ChrisRackauckas:
I think this is fine as an abstraction. It's kind of a detail that Reactant uses Enzyme in order to do its AD. The fact that the Reactant compiler passes use another AD system that is also exposed in Julia is just a detail. To the users, Reactant is just a choice, like ForwardDiff, and it supports some codes and is faster in some cases but has limitations. And importantly, those are different from Enzyme. So to a user of Optimization.jl, SciMLSensitivity.jl, etc,, Reactant is just another choice in the list.

@gdalle if you're fine with the merge I think we just go for it.

gdalle:
I haven't yet read the code, I'll review it today to make sure

docs/src/index.md
```diff
@@ -55,6 +55,12 @@ AutoChainRules
 AutoDiffractor
 ```
 
+### Forward, reverse, or sparse mode
```
gdalle: I recommend a separate category altogether, like "Parallel acceleration", outside of the rest of AD backends. This split would make it clearer that we're talking about a different kind of object.

Also, adding "sparse mode" in there makes it sound like an alternative to the `AutoSparse` struct from ADTypes. But from what I understand, Reactant won't be good at unstructured (scalar-level) sparsity, and will mostly benefit from structured (blockwise) sparsity? This distinction might be confusing for users of the ADTypes ecosystem if not explained.

src/dense.jl
```diff
@@ -82,6 +82,45 @@ function Base.show(io::IO, backend::AutoEnzyme{M, A}) where {M, A}
     print(io, ")")
 end
 
+
+"""
+    AutoReactant{M<:AutoEnzyme}
+
+Struct used to select the [Reactant.jl](https://github.com/EnzymeAD/Reactant.jl) compilation atop Enzyme for automatic differentiation.
```
gdalle:
```suggestion
    AutoReactant{B<:AbstractADType}

Struct used to enforce [Reactant.jl](https://github.com/EnzymeAD/Reactant.jl) compilation atop of the backend performing automatic differentiation.

At the moment, [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl) is the only recommended backend to combine with Reactant.jl (through the [`AutoEnzyme`](@ref) object).

!!! warning
    Since `AutoReactant` is not a standard AD choice, it is not yet supported by [DifferentiationInterface.jl](https://github.com/JuliaDiff/DifferentiationInterface.jl), but it may be supported by other packages.
```
Conceptually it's not tied to Enzyme, as Chris said.
Since DI lives in a different repo, I also want to warn people in the docstring in case they try `AutoReactant` in any DI-enabled workflow and it fails.

gdalle (changes requested):

src/dense.jl
wsmoses: I'm going to push back on this for now as only Enzyme is officailly supported as autodiff during compilation. We may loosen that in the future, at which point we can make a non-breaking change here to loosen the type restriction.

docs/src/index.md
gdalle: I'm very curious then, see https://github.com/EnzymeAD/Reactant.jl/issues/1868

src/dense.jl
gdalle: Fair enough, I understand and  I don't mind having no warning (and I'll try to integrate `AutoReactant` into DI anyway but I don't have a lot of bandwidth right now).

Still, I feel like putting Reactant on top of an AD backend (as opposed to just an Enzyme mode) is conceptually clearer. How about this?

```julia
"""
    AutoReactant{B<:AbstractADType}

Struct used to select the [Reactant.jl](https://github.com/EnzymeAD/Reactant.jl) compilation atop Enzyme for automatic differentiation.

At the moment, [Enzyme.jl](https://github.com/EnzymeAD/Enzyme.jl) is the only recommended backend to combine with Reactant.jl (through the [`AutoEnzyme`](@ref) object).
"""
struct AutoReactant{B <: AbstractADType}
    backend::B
    AutoReactant(backend::AutoEnzyme) = new{typeof(backend)}(backend)
    AutoReactant(backend::AbstractADType) = error("The AD backend $backend is not supported inside `AutoReactant`, try `AutoEnzyme` instead") 
end

AutoReactant() = AutoReactant(AutoEnzyme())
```

wsmoses: I would remain stricter here. Enzyme is the only supported backend. We may augmented this, but I don't want to make any allusions of promises here.

gdalle: Also note that nothing forbids e.g. DI from having an implementation for `AutoReactant{<:AutoForwardDiff}`, except if we force the backend here to be `AutoEnzyme`. Sure it probably won't work as well as for Enzyme, but wouldn't it be interesting for some stuff if Enzyme doesn't have the right rule but Zygote or ForwardDiff does?

I'm not gonna push back further on this. My take is that we're not paying much to add generality and conceptual distinction between AD and Reactant by doing this, but if you really think only Enzyme will work with Reactant then I'll let it be.

wsmoses: If in the future that is supported, we can augment this. But ForwardDiff in particular may be tricky to make work (and definitely is not supported atm).

gdalle: We could do

```julia
struct AutoReactant{B <: AbstractADTypes}
    backend::B
    AutoReactant(backend::AutoEnzyme) = new{typeof(backend)}(backend)
    AutoReactant(backend::AbstractADType) = error("not supported")
    AutoReactant(args...; kwargs...) = AutoReactant(AutoEnzyme(args...; kwargs...))
end
```
which for the end user has the same effect but allows adding stuff later without breaking the API

wsmoses: We can add alternate flags, features, or modes to support that.

Part of why I'm pushing back on this rather strongly is that Reactant provides access to an AD tool/mode inaccessible otherwise.

That AD tool is also built off of Enzyme, but notably _not_ Enzyme.jl. So the constraints or performance implications within Enzyme.jl regarding runtime activity (this does not existing in Reactant)/GC objects/type stability/etc do not apply to accessing Enzyme from within Reactant. Because they have the same basis, non-julia specific Enzyme modes, and other options equally apply.

In principle we can also ingest other code from AD tools as if it were the regular primal code being compiled, as long as it is Reactant-compliant. Notable that means (at the moment) arrays must contain only Reactant primitive types. E.g. not a forwarddiff dual (though you could have a forwarddiff.dual of reactant numbers).

I consider this type more critical for the former at the moment than the latter, though of course we can find a way to support both in the future.

src/dense.jl
```diff
@@ -82,6 +82,45 @@ function Base.show(io::IO, backend::AutoEnzyme{M, A}) where {M, A}
...
+    mode::M
+end
+
+function AutoReactant(;
+        mode::Union{AutoEnzyme,Nothing} = nothing)
+    if mode === nothing
+        mode = AutoEnzyme()
+    end
+    return AutoReactant(mode)
+end
```
gdalle:
```suggestion
```

src/dense.jl
```diff
@@ -82,6 +82,45 @@ function Base.show(io::IO, backend::AutoEnzyme{M, A}) where {M, A}
...
+
+# Constructors
+
+    AutoReactant(; mode::Union{AutoEnzyme,Nothing}=nothing)
+
+# Fields
+
+  - `mode::M` specifies the Enzyme mode of differentiation
+
+      + an object subtyping `EnzymeCore.Mode` (like `EnzymeCore.Forward` or `EnzymeCore.Reverse`) if a specific mode is required
```
gdalle: This is the part that led me astray, and seems inconsistent with `mode::AutoEnzyme`

wsmoses:
@gdalle you all good?

src/dense.jl
```diff
@@ -82,6 +82,45 @@ function Base.show(io::IO, backend::AutoEnzyme{M, A}) where {M, A}
...
+
+# Constructors
+
+    AutoReactant(; mode::Union{AutoEnzyme,Nothing}=nothing)
+
+# Fields
+
+  - `mode::M` specifies the parameterization of differentiation.  It can be:
+
+      + an AutoEnzyme object if a specific mode is required
```
gdalle:
```suggestion
      + an [`AutoEnzyme`](@ref) object if a specific mode is required
```

PR #158, opened 2026-06-14
Canonicalize tests around @safetestset (isolation, matches OrdinaryDiffEq)

ChrisRackauckas-Claude:
## Summary

Canonicalizes the test suite to run each independent Core test unit in its own module via `@safetestset`, matching the canonical OrdinaryDiffEq structure (cross-unit isolation + world-age safety). SafeTestsets expands `@safetestset "name" begin ... end` into a fresh module, so each unit body must be self-contained.

This is the **test-isolation** branch only. It does **not** touch the GROUP-dispatch ladder, `run_tests` harness, or any assertion/logic (the harness migration is a separate branch).

## What changed

- `test/runtests.jl`: each Core unit `@testset "X" begin include("x.jl") end` -> `@safetestset "X" begin include("x.jl") end` (Dense, Sparse, Symbols, Legacy, Miscellaneous, Public). The outer `@testset verbose = true "ADTypes.jl"` wrapper and the `GROUP == "All"/"Core"/"QA"` dispatch are unchanged. QA is untouched.
- `test/test_setup.jl` (new): holds the shared `using` imports and the shared structs/helpers that previously lived at `runtests.jl` top-level — `CustomTag`, `ForwardRuleConfig`/`ReverseRuleConfig`/`ForwardOrReverseRuleConfig`, `FakeSparsityDetector`, `FakeColoringAlgorithm`, `every_ad()`, `every_ad_with_options()`. Each unit file now `include("test_setup.jl")` so its `@safetestset` module is self-contained.
- `test/{dense,sparse,legacy,misc}.jl`: add `include("test_setup.jl")` at the top (symbols.jl/public.jl already carried their own `using ADTypes; using Test`).
- `Project.toml`: add `SafeTestsets` to `[extras]`, `[targets].test`, and `[compat]` (`"0.1, 1"`).

## Subtlety worth flagging (load-bearing)

`using ADTypes` is kept at `runtests.jl` top level (Main) on purpose. Julia qualifies a type's module in `show` based on visibility in `Base.active_module()` (which is `Main`), not the module the test actually runs in. With the units now in fresh modules, `Main` would no longer have ADTypes in scope, so `string(AutoForwardDiff())` becomes `"ADTypes.AutoForwardDiff()"` and the `startswith(string(ad), "Auto")` assertions in the Printing block fail. Keeping `using ADTypes` in Main restores the exact printing context the original tests relied on. (Caught by the local runtime gate — first run had 25 Printing failures, fixed by this.)

## Verification (local, Julia 1.11)

- `GROUP=Core`: `ADTypes.jl | 507 passed, 0 failed` (same 507 tests as before: previously 482 pass + 25 Printing fail in an intermediate state, now all 507 pass).
- `GROUP` unset (All): `507 passed`.
- `GROUP=QA`: `Quality Assurance | 11 passed` (Aqua `deps_compat` happy with the new `[compat] SafeTestsets`).
- Runic `--check --diff` clean on all changed/new files.

Ignore until reviewed by @ChrisRackauckas.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -17,14 +17,16 @@ ADTypesEnzymeCoreExt = "EnzymeCore"
 ChainRulesCore = "1.0.2"
 ConstructionBase = "1.5"
 EnzymeCore = "0.5.3,0.6,0.7,0.8"
+SafeTestsets = "0.1, 1"
 julia = "1.10"
 
 [extras]
 ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
 EnzymeCore = "f151be2c-9106-41f4-ab19-57ee4f262869"
 Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+SafeTestsets = "1bc83da4-3b8d-516f-aca4-4fe02f6d838f"
 Setfield = "efcf1570-3423-57d1-acb7-fd33fddbac46"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["ChainRulesCore", "EnzymeCore", "Pkg", "Setfield", "Test"]
+test = ["ChainRulesCore", "EnzymeCore", "Pkg", "SafeTestsets", "Setfield", "Test"]
--- a/test/dense.jl
+++ b/test/dense.jl
@@ -1,3 +1,5 @@
+include("test_setup.jl")
+
 @testset "AutoChainRules" begin
     ad = AutoChainRules(; ruleconfig = ForwardOrReverseRuleConfig())
     @test ad isa AbstractADType
--- a/test/legacy.jl
+++ b/test/legacy.jl
@@ -1,3 +1,5 @@
+include("test_setup.jl")
+
 @testset "AutoModelingToolkit" begin
     ad_sparse1 = @test_deprecated AutoModelingToolkit(;
         obj_sparse = true, cons_sparse = false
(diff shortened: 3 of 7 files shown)


gdalle:
@ChrisRackauckas what's the deal with the sequence of Claude-generated PRs changing tests and CI? Seems that's all we're getting in ADTypes lately

ChrisRackauckas:
It's the canonicalization roll out matching the new CI runners. Anant talked about it a bit ago:

https://www.youtube.com/watch?v=CFzfJ7F-6Ks

It started to become a major deal with the codecov v6/v7 issues, julia cache action v3 crashing runners (and effectively being a useless waste of time), etc. so most Julia repos were getting random test reds due to different interactions between runners and the random CI script changes. So I took the time to finish Anant's canonicalization, which is a pretty mechanical set of changes. But the canonicalization also found lots of little inconsistencies and issues with the current CI setup when rolled out across the mass of repos (which was the point of course), so the CI scripts needed a little bit in order to be hardened to all of the different possible behaviors. 

There will be another pass that happens hopefully later today that rips out the bespoke runtests.jl and instead uses the new SciMLTesting.jl repo in order to further reduce the repeated code. That will happen right on this PR though, no merging of this until that is done, and with the rollout we'll need to see if there's any edge case missed in SciMLTesting.jl.

While this repo is relatively simple so most of the PRs passed right through it without hitting any troubles, it still will help a lot that it's canonicalized so that for example, the next time CodeCov decides to do a breaking v8, that gets fixed in 1 PR instead of 192, which was the main motivation to finally finish this project.

ChrisRackauckas-Claude:
superseded by the v1.2 folder conversion on sciml-testing-rollout (#159)

PR #172, opened 2026-07-29
Document sparsity pattern element types instead of enforcing Bool

ChrisRackauckas-Claude:
## Summary

Tested float (and other) known sparsity patterns end to end against DifferentiationInterface.jl + SparseMatrixColorings.jl. They all work, so this PR no longer enforces anything — the type restriction is dropped and the observed contract is documented and tested instead. No behavior change, no breaking change.

- document in the `AbstractSparsityDetector` extension contract that consumers must treat every nonzero entry as a structural nonzero rather than requiring `Bool`, that the pattern element type does not propagate to the differentiation result, and that for sparse matrices the *stored* structure defines the pattern (an explicitly stored zero counts as a structural nonzero and gives a valid but more conservative coloring)
- document in both `Known*SparsityDetector` docstrings that `AbstractMatrix{Bool}` is canonical but the element type is neither converted nor checked, and the pattern is handed back exactly as given
- test the pattern element types: `Bool`, `BitMatrix`, `Int`, `UInt8`, `Float64`, `Float32`, `SparseMatrixCSC{Float64}`, `Matrix{Any}` — construction, identity round-trip through `jacobian_sparsity`/`hessian_sparsity`, and `DimensionMismatch` for `Bool`/`Int`/`Float64` patterns
- correct coloring API docstrings to reference `AbstractColoringAlgorithm`
- add `SparseArrays` as a test dependency (for the sparse pattern case)

## Downstream evidence (ADTypes dev from this branch, DI 0.7.20, SparseMatrixColorings 0.4.27)

`AutoSparse(AutoForwardDiff(); sparsity_detector=Known*SparsityDetector(S), coloring_algorithm=GreedyColoringAlgorithm())`, comparing against the dense ForwardDiff result:

```
--- KnownJacobianSparsityDetector through DI.jacobian / DI.jacobian (in-place f!) ---
Matrix{Bool}               -> CORRECT  eltype=Float64
BitMatrix                  -> CORRECT  eltype=Float64
Matrix{Int}                -> CORRECT  eltype=Float64
Matrix{UInt8}              -> CORRECT  eltype=Float64
Matrix{Float64}            -> CORRECT  eltype=Float64
Matrix{Float32}            -> CORRECT  eltype=Float64
SparseMatrixCSC{Bool}      -> CORRECT  eltype=Float64
SparseMatrixCSC{Float64}   -> CORRECT  eltype=Float64
Matrix{Any}                -> CORRECT  eltype=Float64

--- KnownHessianSparsityDetector through DI.hessian ---   (same nine cases, all CORRECT)

--- pattern element type does not leak into the result ---
pattern eltype Bool/Int64/Float32/Float64 -> result eltype Float64

--- sparse pattern with an explicitly stored zero ---
stored zero: correct=true colors=[1, 2, 3]
dropzeros:   correct=true colors=[1, 1, 2]
```

Also probed with a custom detector on registered ADTypes 1.22.3: float patterns with non-0/1 values, negative values, a `NaN` entry, `Complex{Float64}`, and `Union{Missing,Bool}` all produce correct Jacobians. The only failure found is an element type with no `zero`/`iszero` (`Matrix{String}` → `MethodError: no method matching zero(::String)`), which fails in the consumer regardless of what ADTypes accepts.

## Validation
- `julia +1.12 --project=. -e 'using Pkg; Pkg.test()'` → all groups pass (`Sparse | 268 268`)
- `GROUP=QA julia +1.12 --project=. -e 'using Pkg; Pkg.test()'` → `QA | 20 20`
- `julia +1.12 --project=docs docs/make.jl` → no doc warnings/errors
- Runic `--check src test` → exit 0

Ignore until reviewed by @ChrisRackauckas.

### DIFF
--- a/src/sparse.jl
+++ b/src/sparse.jl
@@ -14,6 +14,11 @@ must be an `AbstractMatrix{Bool}` with shape `(length(y), length(x))`, where `y`
 form. A Hessian pattern must be an `AbstractMatrix{Bool}` with shape
 `(length(x), length(x))`. Methods for unsupported operations must throw an error;
 they must not return an unrelated pattern.
+
+New detectors should return `Bool` patterns, but consumers should treat any nonzero
+entry as a structural nonzero: patterns with an integer element type are also
+accepted by [`KnownJacobianSparsityDetector`](@ref) and
+[`KnownHessianSparsityDetector`](@ref), and are handed back unchanged.
 """
 abstract type AbstractSparsityDetector end
 
@@ -52,12 +57,17 @@ hessian_sparsity(f, x, ::NoSparsityDetector) = trues(length(x), length(x))
 
 Trivial sparsity detector used to return a known Jacobian sparsity pattern.
 
+`AbstractMatrix{Bool}` is the recommended pattern type, but any integer element type
+is accepted, so that patterns expressed as matrices of ones and zeroes keep working.
+The pattern is returned by [`jacobian_sparsity`](@ref) unchanged, element type
+included. Non-integer element types (e.g. `Float64`) are rejected by the constructor.
+
 # See also
 
   - [`AbstractSparsityDetector`](@ref)
   - [`KnownHessianSparsityDetector`](@ref)
 """
-struct KnownJacobianSparsityDetector{J <: AbstractMatrix{Bool}} <: AbstractSparsityDetector
+struct KnownJacobianSparsityDetector{J <: AbstractMatrix{<:Integer}} <: AbstractSparsityDetector
     jacobian_sparsity::J
 end
 
@@ -84,12 +94,17 @@ end
 
 Trivial sparsity detector used to return a known Hessian sparsity pattern.
 
+`AbstractMatrix{Bool}` is the recommended pattern type, but any integer element type
+is accepted, so that patterns expressed as matrices of ones and zeroes keep working.
+The pattern is returned by [`hessian_sparsity`](@ref) unchanged, element type
+included. Non-integer element types (e.g. `Float64`) are rejected by the constructor.
+
 # See also
 
   - [`AbstractSparsityDetector`](@ref)
   - [`KnownJacobianSparsityDetector`](@ref)
 """
-struct KnownHessianSparsityDetector{H <: AbstractMatrix{Bool}} <: AbstractSparsityDetector
+struct KnownHessianSparsityDetector{H <: AbstractMatrix{<:Integer}} <: AbstractSparsityDetector
     hessian_sparsity::H
 end
 
--- a/src/sparse.jl
+++ b/src/sparse.jl
@@ -15,10 +15,18 @@ form. A Hessian pattern must be an `AbstractMatrix{Bool}` with shape
 `(length(x), length(x))`. Methods for unsupported operations must throw an error;
 they must not return an unrelated pattern.
 
-New detectors should return `Bool` patterns, but consumers should treat any nonzero
-entry as a structural nonzero: patterns with an integer element type are also
-accepted by [`KnownJacobianSparsityDetector`](@ref) and
-[`KnownHessianSparsityDetector`](@ref), and are handed back unchanged.
+New detectors should return `Bool` patterns, but consumers must not require `Bool`:
+they should treat every nonzero entry as a structural nonzero. Patterns given as
+integer or floating point matrices of ones and zeroes are common in the wild, are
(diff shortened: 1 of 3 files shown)


src/sparse.jl
```diff
@@ -48,7 +48,7 @@ jacobian_sparsity(f!, y, x, ::NoSparsityDetector) = trues(length(y), length(x))
 hessian_sparsity(f, x, ::NoSparsityDetector) = trues(length(x), length(x))
 
 """
-    KnownJacobianSparsityDetector(jacobian_sparsity::AbstractMatrix) <: AbstractSparsityDetector
+    KnownJacobianSparsityDetector(jacobian_sparsity::AbstractMatrix{Bool}) <: AbstractSparsityDetector
```
ChrisRackauckas: @gdalle is this the actual contract you're expecting?

gdalle: In theory yes, in practice I think people passing matrices of (integer) ones and zeros might exist, and the code probably works fine (untested), so I'm not sure it's worth breaking that?

ChrisRackauckas-Claude: Expanded rather than broken: the detectors now accept any integer element type (`Bool <: Integer`), so `Int`/`UInt8`/`Int32` matrices of ones and zeroes keep working and are returned unchanged, element type included. Only non-integer patterns (e.g. `Float64`, `Any`) are rejected.

Documented in both docstrings and in the `AbstractSparsityDetector` extension contract (new detectors should return `Bool`; consumers should treat any nonzero entry as a structural nonzero), and tested: construction + `jacobian_sparsity`/`hessian_sparsity` round-trip over `Bool`, `BitMatrix`, `Int`, `UInt8`, `Int32`, the `DimensionMismatch` paths for integer patterns, and `MethodError` for float/`Any` patterns.

Pushed in https://github.com/SciML/ADTypes.jl/pull/172/commits/cb39d46

ChrisRackauckas-Claude: Tested it downstream rather than guessing. Float patterns work fine, so I dropped the type restriction entirely — the PR is now docs + tests only, no behavior change.

With ADTypes dev'd from this branch, DI 0.7.20, SparseMatrixColorings 0.4.27, `AutoSparse(AutoForwardDiff(); sparsity_detector=Known*SparsityDetector(S), coloring_algorithm=GreedyColoringAlgorithm())` versus dense ForwardDiff:

```
Matrix{Bool} / BitMatrix / Matrix{Int} / Matrix{UInt8} / Matrix{Float64} /
Matrix{Float32} / SparseMatrixCSC{Bool} / SparseMatrixCSC{Float64} / Matrix{Any}
  -> CORRECT for both DI.jacobian (out-of-place and in-place) and DI.hessian
pattern eltype Bool/Int64/Float32/Float64 -> result eltype Float64 in every case
```

Probing further with a custom detector: float patterns with non-0/1 values, negative values, a `NaN` entry, `Complex{Float64}` and `Union{Missing,Bool}` are all correct too. The only thing that breaks is an element type without `zero`/`iszero` (`Matrix{String}` → `MethodError: no method matching zero(::String)`), and that breaks in the consumer regardless of what ADTypes accepts — so there is nothing useful for ADTypes to enforce here.

One behavior worth knowing about, now documented: for a sparse pattern the *stored* structure is what counts, so an explicitly stored zero is treated as a structural nonzero — still correct, just a more conservative coloring (`colors=[1, 2, 3]` with the stored zero vs `[1, 1, 2]` after `dropzeros`).

Pushed in https://github.com/SciML/ADTypes.jl/pull/172/commits/4cc97bc

src/sparse.jl
```diff
@@ -14,6 +14,19 @@ must be an `AbstractMatrix{Bool}` with shape `(length(y), length(x))`, where `y`
...
+integer or floating point matrices of ones and zeroes are common in the wild, are
+accepted unchanged by [`KnownJacobianSparsityDetector`](@ref) and
+[`KnownHessianSparsityDetector`](@ref), and are handled correctly by the coloring
+and decompression implementations used downstream (SparseMatrixColorings.jl through
+DifferentiationInterface.jl). The element type of the pattern does not propagate to
+the differentiation result.
+
+For a sparse matrix, the stored structure defines the pattern: an explicitly stored
+zero is treated as a structural nonzero, which yields a valid but more conservative
+coloring than the same matrix passed through `dropzeros`.
```
ChrisRackauckas, line 29: @gdalle how about this contract?

gdalle: LGTM

PR #176, opened 2026-08-08
AutoHyperHessians: add simd and jet options for HyperHessians 0.3

KristofferC:
HyperHessians 0.3 gained a `simd` config option forcing SIMD.Vec
arithmetic and an explicit opt-in `Jet` representation computing a full
Hessian in one evaluation. Expose both as Bool fields on
AutoHyperHessians so DifferentiationInterface can forward them. They
stay fields rather than type parameters: unlike chunksize they never
need to reach the type domain at the ADTypes level, and a Bool branch
union-splits where an arbitrary Int cannot.


gdalle (approved):
thanks!

PR #177, opened 2026-08-14
Structtypes ext

dingraha:
## Checklist

- [x] Appropriate tests were added
- [x] Any code changes were done in a way that does not break public API
- [x] All documentation related to code changes were updated
- [x] The new code follows the
  [contributor guidelines](https://github.com/SciML/.github/blob/master/CONTRIBUTING.md), in particular the [SciML Style Guide](https://github.com/SciML/SciMLStyle) and
  [COLPRAC](https://github.com/SciML/COLPRAC).
- [x] Any new documentation only uses public API
  
## Additional context

I'm on a quest to be able to serialize/deserialize ADTypes to and from JSON. This PR is the first step. It adds a number of extensions that use StructTypes.jl to handle the logic for constructing the metadata that is eventually passed to JSON3 for the (de)serialization.

This PR makes it possible to serialize most of the ADTypes structs, but not some of the more complicated examples like the AutoSparse backend. That will require similar additions to SparseConnectivityTracer, SparseMatrixColorings, and DifferentiationInterface. That work is on the `structtypes_ext` branches of my forks of those packages, and depends on this PR.

Full disclosure: most of this code was generated by Claude.

PR #180, opened 2026-08-21
Structutils ext

dingraha:
## Checklist

## Additional context

This PR adds serialization/deserialization to/from JSON using StructUtils.jl and JSON.jl. This is an updated version of https://github.com/SciML/ADTypes.jl/pull/177, but uses StructUtils.jl and JSON.jl instead of the deprecated JSON3.jl and less-favored StructTypes.jl. It also moves the StructUtils.jl+JSON.jl functionality for DifferentationInterface.jl, SparseConnectivityTracer.jl, and SparseMatrixColorings.jl from extensions to this package to extensions for those particular packages (see https://github.com/dingraha/DifferentiationInterface.jl/tree/structutils_ext, https://github.com/dingraha/SparseConnectivityTracer.jl/tree/structutils_ext, https://github.com/dingraha/SparseMatrixColorings.jl/tree/structutils_ext).


dingraha:
Also I neglected to say that my buddy claude assisted with this PR. He seems to really like long comments. I will clean those up if needed.

dingraha:
Gentle bump on this. :-)

ChrisRackauckas:
Well first of all it needs a rebase. But secondly, is this even the place for it?

src/ADTypes.jl
```diff
@@ -33,6 +33,35 @@ include("sparse.jl")
...
+    write_ad(ad::AbstractADType) -> String
+
+Serialize an [`AbstractADType`](@ref) to a JSON string.
+
+The output always contains a `"type"` key with the concrete type name, enabling
+round-trip deserialization with [`read_ad`](@ref).
+
+Requires `StructUtils` and `JSON` to be loaded (via the `ADTypesJSONExt` extension).
+"""
+function write_ad end
```
ChrisRackauckas, line 54: why just JSON?

ChrisRackauckas: Why not just use Base.Serialize?

ChrisRackauckas:
4 extensions? This doesn't seem like a good way to do it.

src/ADTypes.jl
dingraha: I think Base.Serialize writes to a binary format, but I need JSON for my application. If we want/need both JSON and Base.Serialize I would be willing to work on that too, though!

dingraha:
> 4 extensions? This doesn't seem like a good way to do it.

I am open to other approaches. :-) My understanding is that using the extensions allows us to keep dependencies out of ADTypes.jl, which appears to be one of the goals of ADTypes.jl. The other approach would be a separate package, I guess? Would that be better?

gdalle:
I think a separate package might make more sense. First, it would not require conditional code loading through extensions. Second, it would allow you to be much freer when designing and updating the serialization. Users could rightfully say that any change in the string associated to a given backand would be breaking, and a breaking ADTypes release would be a burden on the entire ecosystem to update. Instead, keeping ADTypes minimal but having a (much less often used) ADTypesSerialization package that regularly changes, that seems more reasonable to me.

ChrisRackauckas:
Agreed. Also, the fact that you have stacked extensions just shows it's not really an extension. The user needs to add EnzymeCore, JSON, StructUtils in order for certain code to work... that's not really an extension, that's a package with multiple dependencies that have to be managed.

dingraha:
Thanks for the feedback. I'll close this PR and move the code to a new package.

PR #185, opened 2026-09-08
Fix AutoChainRules{Int} assertion on 32-bit Julia

ChrisRackauckas-Claude:
## Summary
- x86 CI fails `ADTypes.Auto(:ChainRules, 1) isa AutoChainRules{Int64}`
- Use `AutoChainRules{Int}` so the type matches native `Int` on both word sizes

## Test plan
- [ ] Core x86 green
- [ ] Existing Core x64 green

Made with [Cursor](https://cursor.com)

### DIFF
--- a/test/symbols.jl
+++ b/test/symbols.jl
@@ -1,7 +1,7 @@
 using ADTypes
 using Test
 
-@test ADTypes.Auto(:ChainRules, 1) isa AutoChainRules{Int64}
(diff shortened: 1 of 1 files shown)


ChrisRackauckas-Claude:
x86 Core is green on this PR. Merge when convenient — unblocks SciML 32-bit CI parity (no architecture-specific hard fails).

